PR #46, opened 2020-02-20
Add flipper movement

### LINKED ISSUE #23: Implement rigid body physics for flippers
Just movement physics, no collision yet.

### LINKED ISSUE #33: Implement flipper hit collider
Flippers have their own hit objects.

freezy:
This PR adds collision and rigid body physics to the flippers. Closes #23 and closes #33.

It adds the collider (`FlipperHit`) and a resolver (`FlipperMover`). We also introduce a `FlipperState`, which holds the runtime data, although that might change in the future.

There were a few minor additional changes:

- Added a registry stub that always returns the fallback value. This is used when retrieving the flipper physics properties.
- Unit tests now use [Fluent Assertions](https://fluentassertions.com/) that are much more intuitive and extensive that xUnit's assertion library.
- When importing a table into Unity, a `TablePlayer` script is attached to the root `GameObject`, which runs the physics loop at each frame, listens to shift key presses and triggers `LeftFlipper` and `RightFlipper` flippers if they exist, and applies the new state to the flipper's transform component.

### DIFF
--- a/VisualPinball.Engine.Test/Test/BaseTests.cs
+++ b/VisualPinball.Engine.Test/Test/BaseTests.cs
@@ -1,5 +1,9 @@
 ﻿using NLog;
 using NLog.Targets;
+using VisualPinball.Engine.Game;
+using VisualPinball.Engine.Math;
+using VisualPinball.Engine.VPT.Ball;
+using VisualPinball.Engine.VPT.Table;
 using Xunit.Abstractions;
 
 namespace VisualPinball.Engine.Test.Test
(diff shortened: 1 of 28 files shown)


freezy:
Yes, actually I have written [tests for that in VPX-JS](https://github.com/vpdb/vpx-js/blob/master/lib/vpt/flipper/flipper-hit.spec.ts#L46), and now I'm thinking about it, they should actually already work here.

freezy:
Added ball tests, and even found a bug.

PR #47, opened 2020-02-22
flipper: rendering mesh improved to look more like physic sim model

ravarcade:
Flipper mesh vertices positions and normal vectors are modified to make flipper flat surfaces tangent to base&tip cylinders.


freezy:
Excellent, thanks a lot!

So the tests fail because VP has the same problem and we test against what VP generates:

image

versus

image

Will update tests and merge later!

freezy:
I've rebased your branch to master in order to get rid of the merge commit.

freezy:
Patch was merged into VPX, so just re-exporting the flipper test file fixed the test.

Thanks again @ravarcade!

PR #48, opened 2020-02-28
Flipper rotation with DOTS

ravarcade:


freezy:
Great, that would have taken ages for me to figure out. Cheers! 👍

PR #51, opened 2020-03-06
Make project UPM (Unity Package Manager) compatible

freezy:
The goal of this PR is that authors can easily start working with VPE and don't need to deal with copying DLLs wrong versions.

This adds a `package.json` to the root of the repository which is an NPM-style declaration of a [package for Unity](https://docs.unity3d.com/Manual/CustomPackages.html). In Unity, we can import the local repository (and in the future via git directly) instead of copying DLLs to the Asset folder.

There are a few benefits to this approach:

- We don't need to ship vendor libs from Unity in order to be able to compile, meaning we don't have version conflicts
- We don't need to figure out which of Unity's package DLLs we need to copy - we can simply declare a dependency to let's say `Unity.Entities` and they will be available.
- Save-to-play is faster, because Unity only compiles changes and doesn't need to reload a whole bunch of DLLs
- It's much easier and more clean to integrate into the host application

There are a few drawbacks as well:

- `.meta` files stink, they are all over the place, and they need to be in source control
- Rider complains about name spaces

## Structure

When importing into Unity, we produce five assemblies:

- `VisualPinball.Engine.dll` - The core VPE engine
- `VisualPinball.Resources.dll` - Static resources, which are big, don't change and don't want to be reloaded every time we update the engine
- `VisualPinball.Unity.dll` - Unity runtime implementation
- `VisualPinball.Unity.Editor.dll` - Unity editor implementation
- `VisualPinball.Unity.Patcher.dll` - Various table fixes

## Other changes

Per default, Unity grabs *absolutely everything* of *all sub folders* of a package. Which is pretty annoying, because we absolutely don't want it to grab everything. For instance, we don't want it grab the build folders of the VS project, or the test fixtures. So instead of using an .ignore file with glob patterns like everybody else, Unity has [some weirdo "Hidden Assets" rules](https://docs.unity3d.com/Manual/SpecialFolders.html). Thus:

- The VS projects now compile into a `.build` folder, which is cleaner anyway
- All test fixtures are now in one folder called `Fixtures~` (yes, with a tilde)

Otherwise:

- The Visual Studio solution now only contains the core projects (no Unity)
- Unlike before, runtime and editor code are now completely separated
- Vendor DLLs from NuGet are now included in the repo, making it (theoretically) possible to load the package via git into Unity.

## Plan B

In order to avoid the `.meta` file plague and keep the core project completely out of Unity, we also thought about moving `package.json` to the Unity project folder and include the core DLLs with the vendor DLLs. The drawback with we see with this approach is that a) we'll need two IDEs open in order to work on both core and Unity at the same time and b) building is slower.

We'll see how it goes, how terrible the `.meta`s are (it'll be pretty bad, specially when refactoring), and might choose that approach later.


freezy:
For those who checked out this repo before this PR went in, make sure to delete the root `package` folder, as well as all `obj` and `bin` folders.

PR #53, opened 2020-03-09
Scripting Setup

freezy:
This PR introduces the notion of an API for the scripting engine, implements an event emitter from a system, and re-organizes the file structure of the Unity project. It also adds an additional Visual Studio Solution that includes the Unity projects, inclusively dependencies.

## API

For the scripting engine we'll need an object tree that implements the scripting API. For this, we've added a `TableApi` and `FlipperApi` class that get instantiated when the table runs.

In order to actually run it, in Unity, create a script that extends `VisualPinballScript`, and attach it to the table's root element. You then have access to the API like that:

```cs
class MyScript : VisualPinballScript
{
	public override void OnAwake(TableApi table)
	{
		table.Flipper("LeftFlipper")?.RotateToEnd();
	}
}
```

## Events

Visual Pinball is heavily based on events. In our API, we use C# events to reflect that feature. For example, in order to subscribe to a flipper rotate event, you would:

```cs
table.Flipper("LeftFlipper").LimitBos += (sender, args) => {
	Debug.Log("Left flipper hit with speed " + args.AngleSpeed);
};
```

Since the events are often triggered from within a `JobComponentSystem`, we use a [`NativeQueue`](https://docs.unity3d.com/Packages/com.unity.collections@0.0/api/Unity.Collections.NativeQueue-1.html) to push the events to and send them to the API implementation after completing the job.

## Structure

In the Unity project, the files are now organized by feature, and not by file type as before. For example, all flipper related classes can be found at `VisualPinball.Unity.VPT.Flipper`.

## Unity Project

We now have a Visual Studio solution that also includes the Unity projects, and their dependencies. There were a few reasons to do that, namely:

- We can now open all files without having to create a Unity project (or even installing Unity )
- Rider just works better. Git diffs are back, the namespaces are correct, and it seems faster too.

The goal is not to compile binaries, but provide a better IDE experience.


ravarcade:
I tested script examples too. Works.
If someone want to test, dont forgot to add using directives to script.
```
using  VisualPinball.Unity.Game;
using VisualPinball.Unity.VPT.Table;
```

PR #56, opened 2020-04-01
custom flipper rotation gizmo

rbxnk:


freezy:
This is awesome! I also like the "pick fix", having to go to the hierarchy and select the parent because it would select just he rubber or the base when clicking on it drove me crazy.

There's just one thing: Try to rotate the top-parent (typically `Table1`) to -98° instead of -90° on the x-axis. Now the gizmo still rotates on the same axis. So it seems it doesn't take into account the parent's transformation?

rbxnk:
Ah, yes. Fixed to rotate around flippers local axis.

freezy:
Merging this, thanks again! If you have time, we'd need the same for transformation, i.e. only allow the X and Y-axes to be used. The Z-axis is determined by which surface the flipper is set to.

PR #57, opened 2020-04-04
Editor/flippers

rbxnk:
2-axis move handle for flippers, couldn't get it to look exactly like the built in, but its pretty close


freezy:
Nice!

However Unity (and Rider) complain about `FinishMove()`'s `newPos` parameter conflicting with the one below. Moving it to up into the class solves the problem.

I'll add that fix, rebase it to master and force-push it to your branch, if that's okay for you.

freezy:
Also cleaned it up a little. Let me know if that's okay for you and I'll merge!

PR #63, opened 2020-05-31
Duplicate project guid fix

jsm174:
When trying to open `VisualPinball.Engine.Unity.sln` with Visual Studio, the following errors are displayed:

`Invalid solution. There are two projects with the same GUID. The project ` ... `VisualPinball.Unity.Patcher.csproj will be ignored.`

`Invalid solution. There are two projects with the same GUID. The project ` ... `VisualPinball.Unity.csproj will be ignored.`

I generated unique guids for these projects and updated the `sln` files accordingly.

Finally, I also added the `VisualPinball.Resources` project as it was a dependency for `VisualPinball.Engine`

Note, this PR still does not allow successful compiling on macOS, as there are still assembly errors:

`
/Users/jmillard/git/jsm174/VisualPinball.Engine/VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallBehavior.cs(13,13): Error CS0012: The type 'Bounds' is defined in an assembly that is not referenced. You must add a reference to assembly 'UnityEngine.CoreModule, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'. (CS0012) (VisualPinball.Unity)`


freezy:
Thanks! :)

PR #67, opened 2020-06-04
First draft for DragPoints edition 

Vroonsh:
- Create IDragPointEditable to expose needed data for edition from Behaviors
- Create DragPointEditor integrated to the existing CustomEditor to handle Dragpoints
- Support Ramps, Rubbers, Surfaces, Triggers (offset issue)
- Draw & Manage Handles for Dragpoints positions (not yet supporting axes limitation)
- Render Centercurve using internal VPE CatmulSpline factory
- Support Undo/Redo for Dragpoints edition (except mesh rebuild)
- Automatically rebuild mesh while editing (some issues with certain meshes)


freezy:
Glanced over it, and it looks pretty nice! @rbxnk will surely have more comments, but for now, can you:

- [ ] Rebase to `master`
- [ ] Get rid of the added `.meta` files - only those of new files should be added, the others are for your build folders and I will put them into `.gitignore`.

VisualPinball.Unity/VisualPinball.Unity/Extensions/MathExtensions.cs
```diff
@@ -24,6 +24,11 @@ public static Vector3 ToUnityVector3(this Vertex3D vertex)
 			return new Vector3(vertex.X, vertex.Y, vertex.Z);
 		}
 
+		public static Vertex3D FromUnityVector3(this Vector3 v)
```
rbxnk, line 27: there's already a "ToVertex3D" which does the same, might have been added to mainline after you did your pass here

VisualPinball.Unity/VisualPinball.Unity.Editor/DragPointsEditor/DragPointsEditor.cs
```diff
@@ -0,0 +1,79 @@
...
+				{
+					Undo.RecordObject(bh, "Change DragPoint Position to " + dpos.ToString());
+					dpos = wlMat.MultiplyPoint(dpos);
+					dpos -= offset;
+					dpos.z -= dpoint.CalcHeight;
+					dpoint.Vertex = dpos.ToVertex3D();
+
+					if (editable != null)
+					{
+						editable.MeshDirty = true;
```
rbxnk, line 53: Setting this flag should happen before the undo.recordobject. That's how I solved the mesh rebuilds for the other inspectors, basically the idea is the mesh is "dirty" in the undo stack, so as soon as the object is reverted that state it'll trigger a mesh rebuild.

VisualPinball.Unity/VisualPinball.Unity.Editor/DragPointsEditor/DragPointsEditor.cs
```diff
@@ -0,0 +1,79 @@
...
+					}
+				}
+			}
+
+			if (controlPoints.Count > 3)
```
rbxnk, line 58: Personally I don't think drawing the polyline is necessary if the mesh is rebuilding in real time, but maybe other folks feel different?

freezy:
Yes, listing them in the inspector would be awesome, including a way to add/remove points rapidly.

VisualPinball.Unity/VisualPinball.Unity/Extensions/MathExtensions.cs
Vroonsh: Yup right I forgot to remove it after using the other one.

VisualPinball.Unity/VisualPinball.Unity.Editor/DragPointsEditor/DragPointsEditor.cs
Vroonsh: Was wondering that :)

VisualPinball.Unity/VisualPinball.Unity.Editor/DragPointsEditor/DragPointsEditor.cs
Vroonsh: I was planning to change the edition Ux in the second step, using this curve with selectable spheres on the drapoint. By doing so, i was thinking about being able to pick along the curve/spheres to add/remove points more easily (like the current VPX edition does)

Vroonsh:
I'll probably try the add/remove point along the curve, maybe with a dropdown menu.
I was also thinking about opening in view Gui with selected points to make them editable.

VisualPinball.Unity/VisualPinball.Unity.Editor/DragPointsEditor/DragPointsEditor.cs
rbxnk: Okay fair enough, it might be more intuitive to have the line there for right clicking on as you say

freezy:
I've cleaned it up. @rbxnk I'll let you approve if everything's fine. @Vroonsh let us know if you want to take on drag points in the inspector in this PR or another one.

Vroonsh:
I'll try to find why some meshes are not updating while editing before closing this PR.

Vroonsh:
Woops just realized your last commit removed DragPointsEditor.cs

freezy:
Oops, sorry! Put it back and renamed the folder while on it. Let me know if anything else's missing, I've still your old branch locally.

Vroonsh:
Yup back to normal :)
I've probably found the problem of some meshes not rebuilt.
It's only ItemBehavior without child and it's in ItemBehavior::RebuildMeshes
Checking if Children is null (it's not the case for Ramps and Surfaces), so trying to find Child transforms, but some of them don't have childs.

Vroonsh:
Fixed the mesh rebuilt. Should be enought for this PR.

freezy:
Thanks guys! 👍

PR #73, opened 2020-06-07
Spinner Collisions

### LINKED ISSUE #36: Implement spinner hit collider
Spinners have their own hit objects (actually, they have a few, two circles for the bracket and a special one for the plate).

freezy:
This implements spinner collisions and movement. Closes #36.

Pretty much the same as gates, with some minor changes in the collision and movement code.

### DIFF
--- a/VisualPinball.Engine/VisualPinball.Engine.csproj
+++ b/VisualPinball.Engine/VisualPinball.Engine.csproj
@@ -88,18 +88,6 @@
 		<Reference Include="System.ServiceModel" />
 		<Reference Include="System.Transactions" />
 		<Reference Include="System.Xml" />
-		<Reference Include="Unity.Entities.Hybrid, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
-		  <HintPath>..\..\vpx_import_test\Library\ScriptAssemblies\Unity.Entities.Hybrid.dll</HintPath>
-		</Reference>
-		<Reference Include="Unity.Mathematics, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
-		  <HintPath>..\VisualPinball.Unity\Vendor\.unity\Unity.Mathematics.dll</HintPath>
-		</Reference>
-		<Reference Include="UnityEngine, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
-		  <HintPath>..\VisualPinball.Unity\Vendor\.unity\UnityEngine.dll</HintPath>
-		</Reference>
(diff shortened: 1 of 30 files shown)


VisualPinball.Engine/VisualPinball.Engine.csproj
```diff
@@ -88,9 +88,18 @@
 		<Reference Include="System.ServiceModel" />
 		<Reference Include="System.Transactions" />
 		<Reference Include="System.Xml" />
+		<Reference Include="Unity.Entities.Hybrid, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null">
```
Vroonsh, line 91: Are these assembly reference needed ? some are pointing to vpx_import_test

VisualPinball.Unity/VisualPinball.Unity/VPT/Spinner/SpinnerDisplacementSystem.cs
```diff
@@ -0,0 +1,94 @@
...
+					// 		m_pspinner->FireGroupEvent(DISPID_SpinnerEvents_Spin);
+					// 	}
+					//
+					// } else {
+					// 	if (movementData.AngleSpeed < target) {
+					// 		m_pspinner->FireGroupEvent(DISPID_SpinnerEvents_Spin);
+					// 	}
+					// }
+
+					while (movementData.Angle > 2.0f * math.PI) {
```
Vroonsh, line 80: maybe movementData.Angle = Mathf.Repeat(movementData.Angle, 0.f, 2.f * math.PI) could work also.

VisualPinball.Engine/VisualPinball.Engine.csproj
freezy: You're right this has nothing to do here.

VisualPinball.Unity/VisualPinball.Unity/VPT/Spinner/SpinnerDisplacementSystem.cs
freezy: Hmm I don't see the signature of your example in [`Mathf.Repeat()`](https://docs.unity3d.com/ScriptReference/Mathf.Repeat.html).

But we're in Burst land here anyway, so we can't use `Mathf`, and Unity's Burst-compatible `math` lib doesn't seem to have something similar.

Vroonsh: Yup sorry was Mathf.Repeat(movementData.Angle, 2.f * math.PI), anyway didn't thought about Burst and math lib limitations

PR #86, opened 2020-06-16
Add OSX support using NetVips 1.2.2 + NetVips Native 8.9.2-build2

jsm174:
The PR updates NetVips to 1.2.2 and uses the latest NetVips Native 8.9.2-build.

This will allow support for MacOS by automatically adding the libvips dylibs to the Vendor folder.


jsm174:
> One little additional request though: Can we drop the `.xml` files in the vendor dir? I think you'd just need to remove `$(OutDir)*.xml` from the `AfterBuild` target and ditch them from the repo (and the `.xml.meta`s as well).

Done!

freezy:
Awesome, thanks a lot!

PR #88, opened 2020-06-17
Updated fixture tests to be cross platform.

jsm174:
This PR allows the `VisualPinball.Engine.Test` `xUnit` test cases to run on both Windows and MacOS.

Since `Path.DirectorySeparatorChar` is a `static readonly` we needed to change from `public const` to `public static readonly`.

This PR also adds a basic github workflow, that matches AppVeyor, and shows that the test cases continue to pass on Windows.

### DIFF
--- a/.github/workflows/VisualPinball.Engine.yml
+++ b/.github/workflows/VisualPinball.Engine.yml
@@ -1,0 +1,15 @@
+on: [push, pull_request]
+
+jobs:
+    vpe-win-x64:
+      runs-on: windows-latest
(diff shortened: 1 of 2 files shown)


freezy:
I'll let you adapt this for NUnit. We can switch to the Unity test runner and deal with x64/x86 issues in another PR.

freezy:
That's now covered by #92, right?

jsm174:
> That's now covered by #92, right?

Correct!

PR #89, opened 2020-06-17
Tests: Switch from xUnit to NUnit

freezy:
Unity uses [NUnit](https://nunit.org/), so let's use that.

However, all assertions now use [Fluent Assertions](https://fluentassertions.com/), so we can pretty easily switch back if necessary.

I've also removed all unnecessary dependencies of the test project, and added the Fluent Assertions to the Unity test project as well. That means the Unity test runner works now with the current tests, however for some reason a few tests are still failing due to rounding errors.

Wondering what it does differently than vanilla NUnit.

One gotcha was the path to the fixtures. Unity compiles the test suite into its `Library/ScriptAssemblies` folder of the project, so the previous relative path to the executing assembly didn't work anymore. 

The trick was finally to use `Packages/<packagename>/` as documented [here](https://docs.unity3d.com/Manual/upm-assets.html).


freezy:
I've update the CI to use NUnit, so tests on master currently fail, so let's get this merged fairly soon ;)

VisualPinball.Engine.Test/Test/Fixtures.cs
```diff
@@ -1,62 +1,87 @@
...
+		public static readonly string Jpg = PathHelper.GetFixturePath("test_pattern.jpg");
+		public static readonly string Png = PathHelper.GetFixturePath("test_pattern.png");
+		public static readonly string PngTransparent = PathHelper.GetFixturePath("test_pattern_transparent.png");
+		public static readonly string Hdr = PathHelper.GetFixturePath("test_pattern_hdr.hdr");
 	}
+
+	public static class PathHelper
+	{
+		public static string GetFixturePath(string filename)
+		{
```
jsm174, line 69: Nice!

PR #91, opened 2020-06-18
Update fixture tests to run on MacOS

jsm174:
This fixes `GetTestPath()` to return the correct folder on MacOS and also Windows.

MacOS was returning: 

`file:///Users/jmillard/_/VisualPinball.Engine/VisualPinball.Engine.Test/.build/Debug`

Windows was returning:

`file://\C:\dev\VisualPinball.Engine\VisualPinball.Engine.Test\.build\Debug`

`System.Uri(<path>).LocalPath` returns the correct platform specific path.


freezy:
Thanks!

Issue #45, opened 2020-02-20
Add editor support for drag points


freezy:
Surfaces, rubbers and walls are based on drag points, which are control points of a [Catmull-Rom spline](https://en.wikipedia.org/wiki/Centripetal_Catmull%E2%80%93Rom_spline). The goal is to visually move the drag points, similar to the Visual Pinball editor:

![image](https://user-images.githubusercontent.com/70426/74916152-a6c7dc00-53c5-11ea-9cd7-85f259f526fe.png)

Unity allows customizing the scene view. The first step is to display the drag points as handles, and then make them movable in the editor. On each movement, the drag point data should be updated, and the mesh regenerated (see [`RebuildMeshes()`](https://github.com/freezy/VisualPinball.Engine/blob/master/VisualPinball.Unity/Components/ItemComponent.cs#L27-L48)), so users see the final mesh while dragging the control points.

There is a short tutorial [here](https://blog.theknightsofunity.com/rendering-custom-gui-scene-view/), API reference is [here](https://docs.unity3d.com/ScriptReference/Editor.OnSceneGUI.html).

If you actually want to draw the spline like in the tutorial (which would be awesome), [here](https://gist.github.com/njvack/6925609) is a snippet that converts Catmull parameters into Bézier, so you can directly use [`DrawBezier()`](https://docs.unity3d.com/ScriptReference/Handles.DrawBezier.html). Or you can [draw them as lines](https://www.habrador.com/tutorials/interpolation/1-catmull-rom-splines/).

Lastly, the drag point parameters should editable in the inspector panel as well. This is somewhat necessary, because a drag point has not only a position, but a few other parameters, such as `IsSlingshot` and `IsSmooth`. I'd say the best UI would be to not render any drag point data in the inspector at all if no drag point is selected, and only the selected drag point if selected.

PR #93, opened 2020-06-19
Full Drag point Editor support

Vroonsh:
Added all the missing features (which i know ;) ) for drag points edition:
- Global Item Lock system (also affecting TransformInspector)
- InspectorGui for Drag points
- Upgraded curve rendering (include slingshots segments)
- Fixed all drag points handling artefacts (height, flickering)
- Added contrains to drag point edition (2D, 3D) regarding ItemBehavior type
- Added dragpoints feature exposition to IDragPointEditable (smooth, slingshot, texture)
- Created HandlesUtils & DataFieldUtils to be reused by other custom editors
- Added drag points flip support

Should close https://github.com/freezy/VisualPinball.Engine/issues/45


VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/DataFieldUtils.cs
```diff
@@ -0,0 +1,226 @@
...
+				{
+					Undo.RecordObjects(recordObjs.ToArray(), $"{message}");
+				}
+				field = val;
+				return true;
+			}
+			return false;
+		}
+
+		public static UnityEngine.Object ItemObjectField(string label, UnityEngine.Object field, bool allowSceneObject, OnEditionChangeDelegate onChange, params (string, object)[] pList)
```
rbxnk, line 176: we need to still pass in the explicit type here, since the "field" var could be null

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/HandlesUtils.cs
```diff
@@ -0,0 +1,74 @@
...
+using UnityEngine;
+using VisualPinball.Unity.VPT;
+
+namespace VisualPinball.Unity.Editor.Utils
+{
+	class HandlesUtils
+	{
+		public delegate void OnPositionChange(Vector3 newPosition, params (string,object)[] plist);
+
+		public static void HandlePosition(Vector3 position, ItemDataTransformType type, Quaternion rotation, OnPositionChange onChange, params (string, object)[] plist)
```
rbxnk, line 11: I'm not a fan of this loose params pattern, I'd much prefer normal params, or at the very least a strongly typed params object.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
+				}
+			}
+
+			[MenuItem(CONTROLPOINTS_MENUPATH + "/IsSlingshot", true)]
+			private static bool SlingshotValidate(MenuCommand command)
+			{
+				ItemInspector editor = command.context as ItemInspector;
+				if (editor == null || editor.DragPointsEditor == null || editor.DragPointsEditor.IsItemLocked()){
+					return false;
+				}
```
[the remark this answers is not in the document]
freezy: Or even better, as mentioned in my main comment, use an in-between abstract class where the drag points editor is never `null`.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
+
+		public bool IsItemLocked()
+		{
+			IEditableItemBehavior editable = _target as IEditableItemBehavior;
+			if (editable == null)
+			{
+				return true;
+			}
+			return editable.IsLocked;
+		}
```
freezy, line 200: Could be a oneliner:

```cs
private bool IsItemLocked => !(_target is IEditableItemBehavior editable) || editable.IsLocked;
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
+
+		public bool HasDragPointExposition(DragPointExposition dpExpo)
+		{
+			IDragPointsEditable dpeditable = _target as IDragPointsEditable;
+			if (dpeditable == null)
+			{
+				return false;
+			}
+			return (dpeditable.GetDragPointExposition() & dpExpo) != DragPointExposition.None;
```
freezy, line 209: This is... tricky. What's the logic behind `DragPointExposition`? I wouldn't expect boolean operations on enums here. Can this be simplified?

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
+					}
+					EditorGUILayout.EndFoldoutHeaderGroup();
+				}
+			}
+		}
+
+		protected void RebuildControlPoints(IDragPointsEditable dpEditable)
+		{
+			IEditableItemBehavior editable = _target as IEditableItemBehavior;
+			if (editable != null)
```
freezy, line 281: C#'s pattern matching is nice for that:

```cs
if (_target is IEditableItemBehavior editable)
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/HandlesUtils.cs
freezy: Agreed. There's also a lot going on here, seems the only usage of `plist` is to pass it to `onChange`? Can this be simplified?

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
-			Vector3 offset = dpeditable.GetEditableOffset();
-			Matrix4x4 lwMat = bh.transform.localToWorldMatrix;
-			Matrix4x4 wlMat = bh.transform.worldToLocalMatrix;
-
-			switch (Event.current.type)
-			{
-				case EventType.Layout:
+			Matrix4x4 wlMat = Enumerable.Count<(string, object)>(plist, pair => pair.Item1 == "wlMat") > 0 ? (Matrix4x4)Enumerable.First<(string, object)>(plist, pair => pair.Item1 == "wlMat").Item2 : Matrix4x4.identity;
+			Vector3 offset = Enumerable.Count<(string, object)>(plist, pair => pair.Item1 == "offset") > 0 ? (Vector3)Enumerable.First<(string, object)>(plist, pair => pair.Item1 == "offset").Item2 : Vector3.zero;
+			IDragPointsEditable dpeditable = Enumerable.Count<(string, object)>(plist, pair => pair.Item1 == "dpeditable") > 0 ? (IDragPointsEditable)Enumerable.First<(string, object)>(plist, pair => pair.Item1 == "dpeditable").Item2 : null;
```
freezy, line 445: This is incomprehensible for me ;)

Tuples that match by string? Casts to matrix and vector? Does this have to be so generic, or can we get that `plist` better typed somehow?

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/DataFieldUtils.cs
freezy: I would like to get rid of those `params (string, object)[] pList` altogether. @Vroonsh it would be great to have more insights of this. Is there an alternative? How many different types are we talking about that made you passing around `object`s?

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
Vroonsh: Hey realize these where useless as params, i can find them back from inside the delegate. Will remove.

Vroonsh:
Hey there, 
Some more insight about this PR. First , i'm learnin both c# and Unity during this contribution, i'm more a cpp programmer...so probably some misunderstanding on some syntax & designs :).

My first DragPointsEditor design was based on UnityEngine.Editor, didn't know about @rbxnk work on ItemInspectors, so i adapted with my first PR, maybe having now an intermediate ItemInspector class handling Dragpoints could be good.

About HandlesUtils & DataFieldsUtils, because i wasn't an ItemInspector nor wasn't directly connected to TransformInspector, i wanted to extract thos Handles & DataFields helpers as services.
Because they where finally unconnected from Inspectors i was wondering it should make sense to have them apart.

Now about the delegate system, I had to set it up because i cannot rely on the Inspectors default FinishMove & FinishEdit calls, i had to have my own on Dragpointeditor side.

then i realized that some context has to be pass through these Utils system to be used during the delegate (for DataFields it was dirtyMesh).
It could have been a context object passed by the call, or some kind of variadic parameters pattern.
Because I had to transfer a second context parameter (sceneRedraw), i opted for the variadic design, more evolutive.

Didn't find a real c# equivalent to variadic templates, so i firstly used params object[] for these delegates. 
the calls looked more like that :
DataFieldUtils.ItemDataField("Ring Speed", ref _bumper.data.RingSpeed, FinishEdit, false);
and the delegate params dispatch like that
bool dirtyMesh = plist.Count >= 1 ? (bool)plist[0] : true;

The problem with this one was that, as default params, you'll have to provide all values from thr first params to setup the one you want.
So i thought about a property version with tuple (string,object).

Well, I think it make sense to have a common management for at least Handles et DataField edition (including the Undo management) so the tool's Ux stays homogeneous.

Rethinking about it, we could also use these Utils class as parent classes of TransformInspector, ItemBehavior & DragpointEditor and override the callbacks FInishEdit & FinishMove, dunno how variadic context params could be handled (context object or variadic params).

I could have a try if you want to.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
Vroonsh: Each Items using drag points don't expose the same data from them.
Ramps : Smooth, SlingShot
Surfaces : Smooth Slingshot, Texturing data
Rubbers : Smooth
Triggers : Smooth, Slingshot
Because of these combinations i used flags to expose features from the DragPointsEditor instance linked to the Item type.
It could also be some IsUsingSmooth, IsUsingSlingShot & IsUsingTexture calls if you prefer or even something else.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
Vroonsh: didn't know :) will change

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
```diff
@@ -1,522 +1,650 @@
...
+			public Vector3 WorldPos = Vector3.zero;
+			public Vector3 ScrPos = Vector3.zero;
+			public bool IsSelected = false;
+			public bool IsFoldOut = false;
+			public int ControlId = 0;
+			public int Index = -1;
+			public float IndexRatio = 0.0f;
+			public List<Vector3> pathPoints = new List<Vector3>();
+
+			public ControlPoint(ref DragPointData dp, int controlID, int idx, float idxratio)
```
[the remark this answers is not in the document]
Vroonsh: yup, will change

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
freezy: Or just a set with the enums, and use [`Contains`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.contains?view=netframework-4.7.2)? I didn't even know that ramps can do slingshots as well...

freezy:
Hey, no worries at all, I'm also still learning a lot about about C# that I hadn't touched before, so that's a good thing. :) 

I would recommend using an IDE that auto-proposes some of the language sugar as well as variable naming. ReSharper for Visual Studio is great, or even better, JetBrain's own C# IDE, [Rider](https://www.jetbrains.com/rider/) (that's what I'm using). I have an open source license for other projects I'm working on, but I will also mention them in VPE's README. As contributor you might be able to get a license too.

I'm strongly in favor of adding an intermediate class for all the drag point logic. And I don't think we have other large features that would also require an intermediate class and require to inherit from multiple classes (which isn't possible in C#). @rbxnk what do you think?

> Now about the delegate system, I had to set it up because i cannot rely on the Inspectors default FinishMove & FinishEdit calls

Why is that? Because the logic is now in its own (non-inspector) class?

About passing arrays, I'm not familiar with variadic templates, but [looking it up](https://en.cppreference.com/w/cpp/language/parameter_pack), it seems they're quite the same as [C# params](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/params)? Can you maybe point me to the code where you needed access to that context that you didn't have, so it's more clear to me?

Vroonsh:
> I'm strongly in favor of adding an intermediate class for all the drag point logic. And I don't think we have other large features that would also require an intermediate class and require to inherit from multiple classes (which isn't possible in C#). @rbxnk what do you think?
> 
I'll migrate all the code to an intermediate ItemInspector (DragPointsItemInspector ?)

> > Now about the delegate system, I had to set it up because i cannot rely on the Inspectors default FinishMove & FinishEdit calls
> 
> Why is that? Because the logic is now in its own (non-inspector) class?
> 
Yup especially the Handles management which was inside the TransformInspector.

> About passing arrays, I'm not familiar with variadic templates, but [looking it up](https://en.cppreference.com/w/cpp/language/parameter_pack), it seems they're quite the same as [C# params](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/params)? Can you maybe point me to the code where you needed access to that context that you didn't have, so it's more clear to me?

Yup i used params, first as simple params object[]. The problem is that it seems not fully implemented as cpp variadic templates are, especially for strongly typed params in variadic methods.
For now my delegate prototype is still exposing params object[] instead of its real parameters, so i then have to gather params inside this array. That's also why i switched to params (string,object)[] to be able to give some context to the values, which is maybe a little bit too complex i agree.

I'll digg a little bit more to see if there is a way to expose strongly typed variadic params in delegates.
You can see another context i needed in ItemInspector.OnPreInspectorGUI() while checking IsLocked (which i cannot use the DataFieldUtils.ItemDataField call because IsLocked cannot be referenced, used the pattern anyway). There i had to call for a Scene Redraw during the ItemInspector.FinishEdit delegate and didn't want to add parameters to FinishEdit each time a new context has to be passed to the delegate.
Designing Handles & DataField utils as services make them be usable from any kind of objects other than Inspectors, but the information which as to be passed through the system back to client is totally up to the client. 
Well, of course, setting the dirtyMesh and/or redrawScene context inside the ItemInspector before calling each ItemDataField and take it back when the delegate is called is also working, but i'm not really fan of this design either.
Another solution would have to pass a context struct to the call so it can evolve without changing the delegate prototype.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
Vroonsh: > Or just a set with the enums, and use [`Contains`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.hashset-1.contains?view=netframework-4.7.2)? I didn't even know that ramps can do slingshots as well...

Yup will change that.
About ramps i don't know, i just went back to VPX and tried any items to see what's exposed for drag points. Maybe it's already a bug or was never used.
Also, when rendering the Slingshot segments in red on curve display, without taking into account the slingshot property exposure, i saw slingshot segments in Rubbers which should only expose Smooth. 
Or maybe i didn't check te correct way in VPX

Vroonsh:
Just saw that,
image

Do we still need to support these helpers ? I can do it on the curve traveller popup menu i think.
The "Add point" is already there.

Or maybe it is part of another issue, because it seems that this picking menu accumulate all available features from all items under the picking position.

Vroonsh:
Ok I fixed almost all the things and added drag points Flip support.
The only thing left is the HandleUtils & DataFieldUtils system.
I've a simplified version with only params object[] in delegates in local.

As long as i'm now deriving from ItemInspector, the DataField part could be done with virtuals and private variables for context (which i'm not a real fan also :)), but still will be bounded to Inspectors, no common management for EditorWindow for instance.

For the transform part, i'm still using it for drag points edition, i can duplicate the PositionHandles code into my Inspector but i'm pretty sure it could also be helpfull somewhere else. On the other side i still have to extract the Rotation & Scale parts.

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
freezy: Hmm okay, weird. I'll have a look the VPX source to check whether slingshots are used otherwise than on surfaces. Ramps (the non-wire ones) are actually used for all kind of stuff in VPX tables, so it wouldn't surprise me if they supported slingshots.

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/HandlesUtils.cs
```diff
@@ -0,0 +1,75 @@
...
+	class HandlesUtils
+	{
+		public delegate void OnPositionChange(Vector3 newPosition, params object[] plist);
+
+		public static void HandlePosition(Vector3 position, ItemDataTransformType type, Quaternion rotation, OnPositionChange onChange, params object[] plist)
+		{
+			HandlePosition(position, type, rotation, 0.2f, 0.0f, onChange, plist);
+		}
+
+		public static void HandlePosition(Vector3 position, ItemDataTransformType type, Quaternion rotation, float handleSize, float snap, OnPositionChange onChange, params object[] plist)
```
rbxnk, line 16: I think we should treat this method as if it was one of Unity's built in handles, i.e. follow their conventions so surrounding code looks similar, and other folks coming to it later will have less to grok.
So basically I guess what I'm saying is, instead of having a change callback, just return the value out of the func, and leave the change check stuff to the caller.

rbxnk: Now that all the drag point code is part of an ItemInspector, can we just get rid of DataFieldUtils and simplify back to the way it was before?

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
freezy: This is still open, right? It's just in `DragPointsItemInspector` now?

VisualPinball.Unity/VisualPinball.Unity.Editor/Editors/DragPointsEditor.cs
freezy: Yeah actually they are. They will just generate a slingshot hit object, which acts the same as if generated through a surface.

freezy:
I've started going through `DragPointsItemInspector` and realized it's nearly 700 lines of code 😮 

I know that coming from C++ that's probably tiny, but I would like to keep files as small as possible. I've noticed that `DragPointsItemInspector.cs` actually contains a bunch of classes. How about moving those into separate files? I could see a separate `VisualPinball.Unity.Editor.DragPoint` namespace for that.

Then as @rbxnk mentioned, there's still `DataFieldUtils`, which I also thought you were merging with the new inspector. Misunderstanding or still on TODO?

Sorry for being a bit of a pain, but it's a huge PR 😃

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/HandlesUtils.cs
Vroonsh: Yup i removed the DataFieldUtils and removed the callback system from the HandlesUtils. Maybe if some EditorWindow not directly heriting from the ItemInspector need some Datafield edition, we could find a way to avoid code duplication at this moment.

Vroonsh:
Hey @freezy 
I'll try to extract the Curve handles & rendering in another class, it's most of the inspector's code.
The other big part is the MenuItems.

freezy:
Yup, let's refactor when necessary. Cheers!

Vroonsh:
It was a pleasure freezy 👍
Always happy to learn new things.
If I can help on other things I'll be pleased !

Cheers

freezy:
There are [a few more low-hanging editor fruits](https://github.com/freezy/VisualPinball.Engine/issues?q=is%3Aissue+is%3Aopen+label%3Aeditor). Feel free to pick one!

PR #97, opened 2020-06-27
Upgrade to .NET Standard 2.0

jsm174:
This PR upgrades Visual Pinball Engine to use .NET Standard 2.0. 

- All `.csproj` files all have been recreated from scratch, adding dependencies as needed. 
- All packages are specified in the `.csproj` file. This is the new way of doing things since `package.json` is deprecated.
- Added `globalPackagesFolder` key to `nuget.config` so `.packages` folder gets created in main directory 
- All files that get deployed to the `Vendor` folder are now explicitly specified.
- Assembly information that used to be in `Properties/Assembly.Info.cs` now lives in the `.csproj` files
- NetVips has a DLL that conflicts with a DLL that is included with Unity Collections. To get around this, NetVips is repacked using `ILRepack`. This repacked DLL has all of its dependancies internalized. This prevents the conflict with Unity Collections version of `System.Runtime.CompilerServices.Unsafe.dll`
- `Directory.Build.props` files have been placed in each of the solution `.sln` folders. This sets the build and obj directories to `.bin` and `.obj`. This prevents Unity from seeing duplicate DLL's in these directories. 
- The Unity managed DLLs have all been upgraded 
- Some updates to the `.gitignore` file have been made to exclude `.bin` and `.obj` folders.
- Unity meta files were added or removed based on the differences


.github/workflows/VisualPinball.Engine.yml
```diff
@@ -12,5 +12,5 @@ jobs:
              dotnet build -c Release
         - name: Test
           run: |
-             .packages/NUnit.ConsoleRunner.3.11.1/tools/nunit3-console.exe VisualPinball.Engine.Test/.build/Release/VisualPinball.Engine.Test.dll
+             .packages/nunit.consolerunner/3.11.1/tools/nunit3-console.exe VisualPinball.Engine.Test/.bin/Release/net472/VisualPinball.Engine.Test.dll
```
freezy, line 15: That's what I'll need to update the CI with, right?

jsm174: Correct.  It is using github actions now though as well, so maybe appveyor isn't needed?

PR #100, opened 2020-07-01
Add github workflow build / test support for Linux

jsm174:
This PR adds support for building and testing VisualPinball.Engine on Linux.

It should be added after PR #99 .

`libminiz.so.*` and `libvips.so.*` native linux libraries have been added and are now copied to `Vendor`.


freezy:
Looks fine, but needs to be rebased!

jsm174:
> Yay!
> 
> So now we can call it cross-platform? :)

I think so.. Just need to run Unity on Ubuntu and give it a try..

rzr:
Is this documented somewhere ? What about supporting other runtimes (like godot?)

freezy:
Well, Unity is not just a runtime but a game engine, so you can't just switch it out with Godot (they work differently).

You'll need Unity to build, and it goes the same for every platform. For a quick setup check see the install instructions:

https://docs.visualpinball.org/creators-guide/setup/installing-vpe.html

If you're aiming for a dev setup, see here:

https://github.com/freezy/VisualPinball.Engine/wiki/Development-Setup

rzr:
yea I know godot is different, then a port should be done, or at least reuse some generic parts if there are any, I remember there was a discussion somewhere about using godot to load vpx files but I lost it...

freezy:
> then a port should be done

That'd be awesome indeed! Let us know when you make any progress :)

rzr:
well the project i am involved in is more lowtech than what you're doing

https://purl.org/rzr/pinball

But I'll be interested into loading vpx assets , that's why extra libs would be welcome

freezy:
Very cool! I remember seeing that on one of the French forums.

The data layer (i.e. loading and saving VPX files) is indeed de-coupled from Unity. You can even fetch it as a [NuGet package](https://www.nuget.org/packages/VisualPinball.Engine/0.0.1-preview.72).

See [tests](https://github.com/freezy/VisualPinball.Engine/blob/master/VisualPinball.Engine.Test/VPT/Primitive/PrimitiveDataTests.cs#L30) for examples of how the API works.

PR #102, opened 2020-07-03
Inspector wrap up

rbxnk:


VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/HitTargetInspector.cs
```diff
@@ -1,82 +1,108 @@
...
+		private static string[] _targetTypeStrings = {
+			"DropTargetBeveled",
+			"DropTargetSimple",
+			"HitTargetRound",
+			"HitTargetRectangle",
+			"HitFatTargetRectangle",
+			"HitFatTargetSquare",
+			"DropTargetFlatSimple",
+			"HitFatTargetSlim",
+			"HitTargetSlim",
```
freezy, line 25: Those are exclusively used for the dropdown, right? Can we name them less cody? Like "Drop Target: Beveled", "Drop Target: Simple", etc?

freezy: Same for gates, kickers, light states, ramps and triggers.

PR #107, opened 2020-07-08
Updated vendor to have platform specific directories

jsm174:
This updates the `Vendor` directory to contain platform specific directories:

```
linux-x64
osx-x64
win-x64
win-x86
```

`NetVips.dll` and `NetMiniZ.dll` are managed dll's and need to be copied to each of these directories because their corresponding native dlls will be loaded from the same directory.

For this to work properly, each dll has a meta file that controls if it is included or not:

Screen Shot 2020-07-07 at 8 27 23 PM
Screen Shot 2020-07-07 at 8 27 25 PM
Screen Shot 2020-07-07 at 8 27 34 PM
Screen Shot 2020-07-07 at 8 27 28 PM


VisualPinball.Engine.Test/VisualPinball.Engine.Test.csproj
```diff
@@ -15,16 +15,13 @@
   </PropertyGroup>
   <ItemGroup>
     <None Remove="**/*.meta" Visible="false" />
-    <None Remove=".bin/*" Visible="false" />
-    <None Remove=".obj/*" Visible="false" />
-    <None Remove="obj/*" Visible="false" />
```
freezy, line 20: Not needed anymore?

freezy:
Oh, and also: Are you going to add win-x86 to GitHub actions? :)

VisualPinball.Engine.Test/VisualPinball.Engine.Test.csproj
jsm174: Correct. In Rider or VS, I'm not seeing those directories. 

Screen Shot 2020-07-08 at 7 04 33 AM

The only way I see them in Rider is if I have it listed to `File System` instead of `Solution`:

Screen Shot 2020-07-08 at 7 04 53 AM

I think the missing backslash at the end of `.obj` in `Directory.build.props` may have fixed this?

jsm174:
The `.swp` files somehow ended up there in a previous commit, so I removed them. I'm still old school and use `vi` from the command line. I really should add a line to .gitignore.

jsm174:
If we are adding/upgrading only managed dll's, we can use rider/vs instead of manually editing the project file.

If we have are upgrading/adding/removing anything with native dll's, we are still going to have to take them directly from the packages. `$outdir` will only have the libs of the platform your compiling for - ie x86, x64, etc.

freezy:
I see, that's because the package versions are hard coded in there I suppose.

I'll just let you remove the `.swp` files and then I'll merge!

jsm174:
> I see, that's because the package versions are hard coded in there I suppose.
> 
> I'll just let you remove the `.swp` files and then I'll merge!

I'm pretty sure all the swp files were deletes.

freezy:
Oops, you're right!

PR #114, opened 2020-07-13
Fix material data when exporting.

### LINKED ISSUE #111: Fix material export
There are a few material attributes that are incorrectly written, mainly due to multiple attributes being packed into a byte:

Original:

image

Written:

image

freezy:
Material data is somewhat obnoxious to parse and to write, since it's a C struct written into game data, so in order to keep retro-compatibility they had to pack multiple values into the available members. Also, bools and chars seem to be of size four.

Anyway, after hours of debugging I think I finally got this right. Tests included.

This closes #111.


rbxnk:
This is awesome, saves me digging into it - noticed some of these discrepancies in the WIP material editor window.

freezy:
Thanks for the review!

PR #119, opened 2020-07-17
auto set some sane project defaults on import

rbxnk:


rbxnk:
This PR should hopefully give folks a better out of the box experience, especially with regards to the light situation.

Issue #79, opened 2020-06-12
Handle materials in editor


freezy:
It's a bit unclear to me how we'll handle materials, so the main goal of this issue is to open the discussion.

![image](https://user-images.githubusercontent.com/70426/84543063-eb6fde00-acfa-11ea-855f-b5de4e21d443.png)

A material in VPX basically defines which shader to use and how to use it. Textures aren't part of the material, they are assigned to the object individually, however, there are physics properties as well.

Now, in order to fully benefit from Unity's renderer, we need to use Unity's shaders, which are obviously a very different beast.

So the first question would be: If authors use Unity just for authoring with the goal of using the VPX player, do we provide a way to deal with materials? There is no way of providing an accurate visualization, so what we could do is the following:

- When a Unity material is created or edited, create or update a VPX material as well. There are a bunch of parameters such as colors or shininess than can be converted in order achieve a similar look.
- Provide a VPX materials editor clone where an author can fine-tune the VPX materials. Knowing those won't affect Unity's shaders. 
- The VPX materials should be "lockable", i.e. they then won't be updated when corresponding Unity material updates.

In short, authors should be able to apply Unity materials to their meshes without any restrictions, but since we still want to be able to play the produced `.vpx` file in VPX, they should be converted and editable.

How we internally handle all that, which of Unity's render pipelines we support, how we write the Unity materials to the `.vpx` file etc will be part of another issue.

Thoughts?

rbxnk:
My vote would be to do the following:

* Get as close as we can to VPX's look with a set of materials we create in Unity, drive those materials properties solely from the vpx material data.
* Introduce some new VPE only fields for these materials that allow authors to define renderer-specific material properties. Authors might want to tweak the look of something if its pbr vs a more traditional shader, or raytraced. But all this still using materials we provide that takes their values from the VPE item fields.

This would provide enough control to get us rolling, without having to solve if/how we pack in unity assets yet.

freezy:
So the hard requirements from my side would be the following:

- When importing a vanilla `.vpx` file, materials should look somewhat similar in VPE compared to VPX's renderer.
- When creating a table in VPE, materials should also look similar in VPX.
- In VPX, authors should be able to tweak the material without changing the VPE material (given it's already "tweaked").
- In VPE, authors should have full access to all the shaders (incl shadergraph and custom shaders) when working with materials (how we convert those to VPX materials is another question).
- When loading a `.vpx` file into Unity, the used materials should be editable

[...]

PR #122, opened 2020-07-18
Material manager

rbxnk:
This PR is what I'd consider the bare minimum material support for VPE, working towards #79. It includes a material editor window, modeled after the VPX UI, supports editing with live updates, adding removing and exporting.

No Unity-specific material support yet, just getting to VPX parity first.

Capture


VisualPinball.Engine/VPT/Bumper/Bumper.cs
```diff
@@ -8,6 +8,7 @@ public class Bumper : Item<BumperData>, IRenderable, IHittable
 	{
 		public bool IsCollidable => true;
 		public EventProxy EventProxy { get; private set; }
+		public string[] UsedMaterials => new string[] { Data.CapMaterial, Data.BaseMaterial, Data.RingMaterial, Data.SocketMaterial };
```
freezy, line 11: I assume this is called by the editor loop? No issue with garbage collection if we instantiate it each time? Maybe put it on a readonly prop and just return it?

VisualPinball.Engine/Math/Color.cs
```diff
@@ -17,6 +17,14 @@ public class Color
...
 		public int Bgr => Blue * 65536 + Green * 256 + Red;
 
+		public Color(Color other)
+		{
+			Red = other.Red;
+			Green = other.Green;
+			Blue = other.Blue;
+			Alpha = other.Alpha;
+		}
+
```
freezy, line 27: I usually use a `.Clone()` method that returns a copy, but it's fine if you prefer explicitly instantiating it.

VisualPinball.Unity/VisualPinball.Unity.Editor/Materials/MaterialEditor.cs
```diff
@@ -0,0 +1,330 @@
+using System.Collections.Generic;
+using System.Linq;
+using Unity.Mathematics;
```
freezy, line 3: Doesn't seem to be used and we don't use Unity's legacy math lib, so it's red in my IDE

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
```diff
@@ -12,12 +13,15 @@
...
 {
-	public abstract class ItemBehavior<TItem, TData> : MonoBehaviour, IEditableItemBehavior where TData : ItemData where TItem : Item<TData>, IRenderable
+	public abstract class ItemBehavior<TItem, TData> : MonoBehaviour, IEditableItemBehavior
+		where TData : ItemData where TItem : Item<TData>, IRenderable
 	{
 		[SerializeField]
 		public TData data;
 
 		public TItem Item => _item ?? (_item = GetItem());
+		public bool IsLocked { get { return data.IsLocked; } set { data.IsLocked = value; } }
```
freezy, line 23: Detail, but this is nicer:

```cs
public bool IsLocked { get => data.IsLocked; set => data.IsLocked = value; }
```

freezy: Just saw it was there before actually.

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
```diff
@@ -102,7 +112,16 @@ public void RebuildMeshes()
...
 		public virtual void SetEditorScale(Vector3 rot) { }
 
-		public bool IsLocked { get { return data.IsLocked; } set { data.IsLocked = value; } }
+		public virtual void HandleMaterialRenamed(string undoName, string oldName, string newName) { }
+
+		// rename helper to cut down on the boiler plate in the concrete classes
+		protected void TryRenameField(string undoName, ref string field, string oldName, string newName)
+		{
+			if (field == oldName) {
+				Undo.RecordObject(this, undoName);
```
freezy, line 121: My IDE doesn't know `Undo`. Do you know from which DLL it comes from?

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
```diff
@@ -121,6 +140,10 @@ protected virtual void Awake()
 
 		protected virtual void OnDrawGizmos()
 		{
+			if (_meshDirty) {
+				RebuildMeshes();
+			}
+
```
freezy, line 146: Unrelated bugfix I assume?

VisualPinball.Engine/VPT/Bumper/Bumper.cs
rbxnk: No its not called regularly, just when the window's rows need to populate to get the "in use" flag and when the mat gets edited to dirty the items. If we make a persistent field for it we'd have to also add a dirty flag or some update call to rebuild the list when the item changes. I was trying to keep things simple, but let me think on it, see if we can build a simple pattern for all that.

freezy: If it's not called every frame, that's fine then.

VisualPinball.Engine/Math/Color.cs
rbxnk: I don't have a preference either way, i'll convert to clone methods

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
rbxnk: Yeah, just moved it up but i'll change it while I'm in there

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
rbxnk: No actually, this is meant to cover the case where you undo a material change and we need to rebuild meshes for the affected items. We don't want to rely on the inspector windows themselves to handle it (they won't necessarily be visible after all), so instead we'll check for dirty when the scene views draw.

freezy: Gotcha, thanks for the explanation!

VisualPinball.Unity/VisualPinball.Unity/VPT/ItemBehavior.cs
rbxnk: This was another case of editor assembly stuff, I wrapped it in an ifdef to fix. I know its not ideal to have editor code in here (though to be fair all that stuff around there is, even if it doesn't live in that assembly). I'm sure we could refactor this out, but it would add a lot of complexity plumbing results all the back out to the callers etc.

freezy: Okay for me!

PR #128, opened 2020-07-21
layers : real layers structure visualization

Vroonsh:
- Setup TreeView with generic TreeElement types (automatic TreeViewItem construct & generic events handling)
- Upgraded all LayerEditor & TreeView classes to this framework
- Built layers structure base on BiffData from Table items
- Added a search field
- Added layer renaming
- LayerEditor/SceneManagement view Visibility synchronization
- Mixed visibility support for layers & table items
- GameObjects selection through LayerEditor double-click
- Layer Add/Remove
- Undo/Redo Management


VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
```diff
@@ -13,7 +14,12 @@ public class LayerEditor : EditorWindow
 	{
 		private TreeViewState _treeViewState;
 		private TreeView _treeView;
+		private SearchField _searchField;
 
+		private LayerHandler _layerHandler;
+
+		private Rect _toolbarRect { get { return new Rect(20f, 10f, position.width - 40f, 20f); } }
+		private Rect _treeViewRect { get { return new Rect(20, 30, position.width - 40, position.height - 60); } }
```
freezy, line 22: Is that the padding? I would prefer it a bit smaller, it kinda stands out now when the window is docked.

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerHandler.cs
```diff
@@ -1,36 +1,102 @@
...
+				BuildLayersRecursively(child);
 			}
 		}
-		public void SetVisibility(GameObject[] gameObjs, bool visibility, bool includeDescendants)
+
+		private void AddToLayer(Behaviour bh)
+		{
+			if (bh is VPT.ILayerableItemBehavior layerableItemBehavior) {
+				if (layerableItemBehavior.EditorLayerName == "") {
+					layerableItemBehavior.EditorLayerName = $"Layer_{layerableItemBehavior.EditorLayer + 1}";
```
freezy, line 66: Will this update the data as well? I.e. will the layer name be written to the `.vpx` when exporting?

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
freezy: Also, less boilerplate would be:
```cs
private Rect _toolbarRect => new Rect(20f, 10f, position.width - 40f, 20f);
private Rect _treeViewRect => new Rect(20, 30, position.width - 40, position.height - 60);
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
```diff
@@ -26,11 +32,34 @@ protected virtual void OnEnable()
 			if (_treeViewState == null)
 				_treeViewState = new TreeViewState();
 
-			_treeView = new LayerTreeView(_treeViewState);
+			if (_layerHandler== null)
+				_layerHandler = new LayerHandler();
+
+			if (_searchField == null)
+				_searchField = new SearchField();
```
freezy, line 39: I'm not going to nag you anymore about this, but prefer curly brackets even around oneliners. They make diffs easier to read, and are less error-prone.

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
```diff
@@ -0,0 +1,114 @@
...
+	/// Element type used in TreeModel for the LayerTreeView
+	/// </summary>
+	/// <remarks>
+	/// Could be initialized in different ways to fit all possible LayerTreeView item types
+	/// </remarks>
+	class LayerTreeElement : TreeElement
+	{
+		public Table Table { get; private set; }
+		public ILayerableItemBehavior Item { get; private set; }
+		public string LayerName { get; private set; }
```
freezy, line 35: No need for `set` as far as I can see

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
```diff
@@ -0,0 +1,114 @@
...
+		public LayerTreeViewElementType Type { get { return Table != null ? LayerTreeViewElementType.Table : Item != null ? LayerTreeViewElementType.Item : LayerName != null ? LayerTreeViewElementType.Layer : LayerTreeViewElementType.Root; } }
```
freezy, line 36: More readable:

```cs
public LayerTreeViewElementType Type => Table != null 
	? LayerTreeViewElementType.Table 
	: Item != null 
		? LayerTreeViewElementType.Item 
		: LayerName != null 
			? LayerTreeViewElementType.Layer 
			: LayerTreeViewElementType.Root;
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
```diff
@@ -0,0 +1,114 @@
...
+
+		[SerializeField]
```
freezy, line 38: Rider tells me this is redundant. Maybe all members of a `TreeElement` are serialized?

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeElement.cs
```diff
@@ -0,0 +1,49 @@
...
+	[Serializable]
+	public class TreeElement
+	{
+		[SerializeField]
+		public int Id { get; set; }
+		[SerializeField]
+		public virtual string Name { get; }
+
+		[SerializeField]
+		public int Depth { get; set; }
```
freezy, line 24: I think the `[Serializable]` on top of the class makes all members serializable per default, so only the `[NonSerialized]` attributes are needed.

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeElement.cs
```diff
@@ -0,0 +1,49 @@
...
+
+		[NonSerialized]
+		private TreeElement _parent;
+		public TreeElement Parent { get { return _parent; } set { _parent = value; } }
```
freezy, line 28: Any particular reason to make this a property and not just a public (non-serialized) field? Same question for `Children` below

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeModel.cs
```diff
@@ -0,0 +1,232 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+
+namespace VisualPinball.Unity.Editor.Utils.TreeViewWithTreeModel
```
freezy, line 6: Is this layer-agnostic?

I would prefer a shorter name. How about `VisualPinball.Unity.Editor.Utils.TreeView`?

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerHandler.cs
Vroonsh: yup it will

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
Vroonsh: sure :)

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerHandler.cs
freezy: Cool!

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
Vroonsh: removed

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
Vroonsh: done

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
Vroonsh: It was mainly in case of saving TreeModel to Asset afaik

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeElement.cs
Vroonsh: nope except it came from the Unity TreeViewWithModel sample :)

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeModel.cs
Vroonsh: yup it's a base framework to easily populate a TreeModel of TreeElements into a special TreeView called TreeViewWithTreeModel, the TreeView base class is not as easy to setup.

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
Vroonsh: cannot use this syntax with non static fields.

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeViewWithTreeModel/TreeElement.cs
freezy: Yeah some people's idea of OOP is that all fields should be private. However I fail to see the advantage (in most cases, anyway) but only the boiler plate :)

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
freezy: Yes, you can!

What you can't do is make it a field and create the object at class instantiation.

Vroonsh:
> Good stuff! I've mainly looked at your syntax, I'll have a more closer look tomorrow at the logic. Two comments in terms of features so far:
> 
> * Can't move items to other layers
> * Can't select it - double-clicking on a layer should select it in the inspector so the user can quickly edit it
> 
> Can you maybe describe a bit how you've tackled the problem? I see you had to flatten the tree in order to be able to serialize it? Is that a common Unity problem? Why does it need to be serialized in the first place? How often is this done?
> 
> EDIT: I have no problem with the two features in another PR, just keeping track :)

Yup, found this TreeViewWithTreeModel framework interresting firstly because of all the existing helpers provided with it (search, drag n drop, items manipulation within the model, events, model/view design), the serialization wasn't my main need. But yeah, the list format is there especially for serialization purpose.

Vroonsh:
Current visual status of the Layer Editor
image

freezy:
Hey, thanks for the updates! Are you still working on it, or can I give it another review?

Vroonsh:
> Hey, thanks for the updates! Are you still working on it, or can I give it another review?

Yup, got cleaned up the code, should be ok

Vroonsh:
I've still one concern about items layer assignation, for now it's only with drag&drop, would like to also propose a contextual menu on items like "Add to new layer" & "Add to layer" with the existing layer list as submenu. 
But afaik the MenuItems system cannot do dynamic submenu creation. 
Anyway, could be part of future improvements.

Vroonsh:
Ok upgraded the Context menu, way faster to reassign items to layers.
Here how it looks :
image

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerHandler.cs
```diff
@@ -1,35 +1,300 @@
...
+		/// <param name="newName">the new validated name</param>
+		internal void OnLayerRenamed(LayerTreeElement element, string newName)
+		{
+			if (element.LayerName == newName) {
+				return;
+			}
+
+			//Check if there is not already a layers with thr same name
+			if (_layers.ContainsKey(newName)) {
+				EditorUtility.DisplayDialog("Visual Pinball", $"There is already a layer named {newName}.\nFind another layer name.", "Close");
```
freezy, line 172: VPX allows duplicate layer names. I agree it doesn't make much sense in terms of usability, but that means we can get duplicate layer names from imports, which, as far as I can see, will break your `_layers` logic. Maybe use the layer ID to index?

Vroonsh: Right, I'll check that

freezy: Hmm what do you mean by reloading? Save, close and open? How are the layers written?

In general the identifier seems to be the ID, not the name. Why not use the ID then? Does the trigger other problems?

Vroonsh: VPX7 is using LANR data first when loading items. If it's empty it creates a "Layer_num"  layer name using LAYR data and set LANR data with it. On next save, LAYR shouldn't be used anymore.
You can see it at the end of PinTable::LoadGameFromStorage in VPX7 project.
I did the same when initializing layer names when opening the editor.
Tthe LANR item's data is the only place where layer names are stored, so having 2 layers with the same name last time you saved your table will lead to layer merging the next time you'll load it.
VPX6 tables are not using LANR at all, so i'll create "Layer_num" layers in this case (as VPX7 does).

VisualPinball.Unity/VisualPinball.Unity.Editor/Utils/TreeView/TreeElement.cs
```diff
@@ -0,0 +1,75 @@
...
+	/// TreeElement is the base class for creating TreeModel which will be used by TreeViewWithTreeModel, you'll have to inherit from it.
+	/// It provides base properties to be inserted & identified in a TreeView (Id, Name, Depth)
+	/// It hosts also its hierarchical information.
+	/// </summary>
+	public abstract class TreeElement
+	{
+		/// <summary>
+		/// Unique ID of the TreeElement within the Tree structure
+		/// </summary>
+		public int Id { get; set; }
```
freezy, line 17: I'm wondering... do the IDs have to be sequential? If not, maybe using the game object's [instance ID](https://docs.unity3d.com/ScriptReference/Object.GetInstanceID.html) would be easier?

Vroonsh: I'm already using gameobject instance Id for items treeelements, only layers treeelements are using indexes, they're not related to gameobjects

freezy: Ah, you're right, sorry I missed that!

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeElement.cs
```diff
@@ -0,0 +1,209 @@
...
+
+				if (Item is IIdentifiableItemBehavior identifiable) {
+					return identifiable.Name;
+				}
+
+				if (LayerName != null) {
+					return LayerName;
+				}
+
+				return "<Root>";
```
freezy, line 102: Can't we rely on the `Type` here as well (like below)?

Vroonsh: Yup sure you're right !

freezy: Cool, I've pushed a few changes meanwhile. And I'm nearly through!

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerTreeView.cs
```diff
@@ -1,166 +1,145 @@
...
+						ItemsDropped?.Invoke(elements, layerTreeItem.Data);
+					}
+				}
+				Reload();
+				return DragAndDropVisualMode.None;
+			} else {
+				switch (args.dragAndDropPosition) {
+					default: {
+						return DragAndDropVisualMode.None;
+					}
```
freezy: Hmm shouldn't this be at the end?

VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerHandler.cs
freezy: If you're saying VP doesn't save multiple layers with the same name, then it's okay.

Vroonsh:
> I haven't reviewed the `VisualPinball.Unity.Editor.Utils.TreeView` namespace, since to my understanding, it's pretty much what came from the example code from the Unity doc.
> 
In fact there is not so much left from the sample except maybe the Search feature and the TreeElement structure.
Added most of the TreeElement helpers and utility methods
Also refactored the renaming, double/context click & drag & drop.

> Thanks again, this is an awesome feature!
My pleasure :)

Vroonsh:
New '+' button
image

PR #134, opened 2020-07-27
Add kicker collision

freezy:
This PR adds support for the kicker during gameplay. 

There are two classes for dealing with physics: In the broad- and nearphase, a `CircleCollider` is used, while for resolving the collision, `KickerCollider` is used.

In "non-legacy" mode, the kicker comes with an additional collision mesh, which makes the ball more naturally roll into the hole. The vertices of that mesh are converted to a `BlobAssetReference` and put on a separate component that is only set when legacy is disabled. This still needs to be tweaked because it's awfully slow.

## Unrelated changes

- Bump all Unity dependencies to latest
- Convert remaining game objects to entities (ramps and rubbers)
- Create remaining APIs (bumper, ramp)
- Port collision code for poly 3D collider

## TODO

- [ ] ~Fix~ Mitigate performance issue with non-legacy collision
- [ ] ~Add API for ball creation~ Separate PR


freezy:
This has been sitting around for too long, and fixed will go into `editor/toolbox` anyway. Merging.

PR #136, opened 2020-07-31
Dedicated Material classes for Standard, URP and HDRP

Roland09:
Created dedicated material converters and ball material creators for Standard, URP and HDRP graphic pipelines

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
@@ -3,8 +3,8 @@
 using System;
 using System.Text;
 using NLog;
-using UnityEngine;
 using VisualPinball.Engine.VPT;
+using VisualPinball.Unity.Import.Material;
 using VisualPinball.Unity.VPT.Table;
 using Logger = NLog.Logger;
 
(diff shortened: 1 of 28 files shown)


VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
```diff
@@ -3,144 +3,75 @@
...
-				debug?.AppendLine("Metallic set to 1.");
-			}
+		private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
 
-			// roughness / glossiness
-			unityMaterial.SetFloat(Glossiness, vpxMaterial.Roughness);
+		/// <summary>
+		/// Lazy instantiation of the material converter
+		/// </summary>
+		private static readonly Lazy<IMaterialConverter> lazyMaterialConverterInstance = new Lazy<IMaterialConverter>(() => CreateMaterialConverter());
```
rbxnk, line 21: Lazy isn't something I come across much, is there a good reason to use this over simply having an IMaterialConver member and in the property just doing something like `Foo => _foo ?? _foo = new bar()`?

VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
```diff
@@ -3,144 +3,75 @@
...
-			// blend mode
-			ApplyBlendMode(unityMaterial, vpxMaterial.MapBlendMode);
-			if (vpxMaterial.MapBlendMode == BlendMode.Translucent) {
-				col.a = Mathf.Min(1, Mathf.Max(0, vpxMaterial.Opacity));
-				unityMaterial.SetColor(Color, col);
-			}
+		/// <summary>
+		/// Singleton Material Converter
+		/// </summary>
+		public static IMaterialConverter MaterialConverter { get { return lazyMaterialConverterInstance.Value; } }
```
rbxnk, line 26: We've been preferring the => shorthand whenever possible

rbxnk:
This all looks awesome. I do wonder though why we need the separation of the ball materials. I know all that predates this change, but I wonder if we could just route them through the table's mat/factory?

VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
Roland09: Thanks, removed it.

VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
Roland09: Thanks, I updated the code.

VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
```diff
@@ -16,15 +16,9 @@ public static class MaterialExtensions
...
+		/// Material Converter instance for the current graphics pipeline
 		/// </summary>
-		private static readonly Lazy<IMaterialConverter> lazyMaterialConverterInstance = new Lazy<IMaterialConverter>(() => CreateMaterialConverter());
-
-		/// <summary>
-		/// Singleton Material Converter
-		/// </summary>
-		public static IMaterialConverter MaterialConverter { get { return lazyMaterialConverterInstance.Value; } }
-
+		public static IMaterialConverter MaterialConverter => CreateMaterialConverter();
```
rbxnk, line 21: Sorry didn't mean the idea of a singleton was bad, was just curious about the benefits of Lazy over a simple singleton one liner. I think we probably would prefer not to instantiate these very call

freezy:
I don't mind having the ball creation separate. It allows for more flexibility, and since the ball isn't part of the table, we don't need to keep backwards compatibility.

VisualPinball.Unity/VisualPinball.Unity/Extensions/MaterialExtensions.cs
freezy: Reading [this](https://docs.microsoft.com/en-us/dotnet/framework/performance/lazy-initialization) it simply seems a thread-safe way of doing the same thing. If it's part of .NET Standard and supported by Unity I don't mind using it.

PR #140, opened 2020-08-04
Import changes

Roland09:
* Changing gameobject name of the table to the table info name.
* Patching Jurassic Park broken normal map, i. e. removing it. It caused the editor window to become black. It's a Unity thing.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/Import/Material/HdrpMaterialConverter.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/Import/Material/HdrpMaterialConverter.cs
@@ -9,23 +9,6 @@ namespace VisualPinball.Unity.Import.Material
 {
 	public class HdrpMaterialConverter : IMaterialConverter
 	{
-		private readonly int SurfaceType = Shader.PropertyToID("_SurfaceType");
-		private readonly int BaseColor = Shader.PropertyToID("_BaseColor");
-		private readonly int BaseColorMap = Shader.PropertyToID("_BaseColorMap");
-
-		private readonly int NormalMap = Shader.PropertyToID("_NormalMap");
-		private readonly int NormalMapSpace = Shader.PropertyToID("_NormalMapSpace");
-		private readonly int NormalScale = Shader.PropertyToID("_NormalScale");
-
-		private readonly int Metallic = Shader.PropertyToID("_Metallic");
-		private readonly int Smoothness = Shader.PropertyToID("_Smoothness");
-		private readonly int BlendMode = Shader.PropertyToID("_BlendMode");
-		private readonly int TransparentSortPriority = Shader.PropertyToID("_TransparentSortPriority");
-		private readonly int AlphaCutoffEnable = Shader.PropertyToID("_AlphaCutoffEnable");
-		private readonly int SrcBlend = Shader.PropertyToID("_SrcBlend");
-		private readonly int DstBlend = Shader.PropertyToID("_DstBlend");
-		private readonly int ZWrite = Shader.PropertyToID("_ZWrite");
-
 		public Shader GetShader()
 		{
 			return Shader.Find("HDRP/Lit");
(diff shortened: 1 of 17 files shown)


VisualPinball.Unity/VisualPinball.Unity/Import/Material/UrpMaterialConverter.cs
```diff
@@ -9,25 +9,11 @@ namespace VisualPinball.Unity.Import.Material
...
-		private readonly int BumpMap = Shader.PropertyToID("_BumpMap");
-		private readonly int Metallic = Shader.PropertyToID("_Metallic");
-		private readonly int Smoothness = Shader.PropertyToID("_Smoothness");
-		private readonly int Blend = Shader.PropertyToID("_Blend");
-		private readonly int QueueOffset = Shader.PropertyToID("_QueueOffset");
-		private readonly int AlphaClip = Shader.PropertyToID("_AlphaClip");
-		private readonly int SrcBlend = Shader.PropertyToID("_SrcBlend");
-		private readonly int DstBlend = Shader.PropertyToID("_DstBlend");
-		private readonly int ZWrite = Shader.PropertyToID("_ZWrite");
-
```
freezy, line 24: Same note as above. Why remove them?

VisualPinball.Unity/VisualPinball.Unity/Import/VpxImporter.cs
```diff
@@ -59,9 +59,12 @@ public void Import(string fileName, Table table)
 			_table = table;
 
 			var go = gameObject;
-			go.name = _table.Name;
+			
 			MakeSerializable(go, table);
 
+			// set the gameobject name; this needs to happen after MakeSerializable because the name is set there as well
+			go.name = "Table [" + _table.InfoName + "]";
+
```
freezy, line 67: As discussed, as GameObject names are in sync with the data names, this will break the table script.

VisualPinball.Unity/VisualPinball.Unity/Import/Material/HdrpMaterialConverter.cs
```diff
@@ -9,23 +9,6 @@ namespace VisualPinball.Unity.Import.Material
...
-		private readonly int NormalScale = Shader.PropertyToID("_NormalScale");
-
-		private readonly int Metallic = Shader.PropertyToID("_Metallic");
-		private readonly int Smoothness = Shader.PropertyToID("_Smoothness");
-		private readonly int BlendMode = Shader.PropertyToID("_BlendMode");
-		private readonly int TransparentSortPriority = Shader.PropertyToID("_TransparentSortPriority");
-		private readonly int AlphaCutoffEnable = Shader.PropertyToID("_AlphaCutoffEnable");
-		private readonly int SrcBlend = Shader.PropertyToID("_SrcBlend");
-		private readonly int DstBlend = Shader.PropertyToID("_DstBlend");
-		private readonly int ZWrite = Shader.PropertyToID("_ZWrite");
```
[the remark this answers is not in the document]
Roland09: Because they become more and more and the code less readable, i. e. it's better readable with the direct property name. Performance-wise it shouldn't have an effect, since you import a pinball table only once. If it really were a problem, Unity themselves wouldn't do it that way. Just look at StandardsToHDLitMaterialUpgrader.cs, BaseLitGUI.cs, HDLitMasterNode.cs, etc.

VisualPinball.Unity/VisualPinball.Unity/Import/Material/UrpMaterialConverter.cs
Roland09: See above.

VisualPinball.Unity/VisualPinball.Unity/Import/Material/HdrpMaterialConverter.cs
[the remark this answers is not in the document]
freezy: Okay, I had a look at this. [Here](https://github.com/JetBrains/resharper-unity/wiki/Avoid-using-string-based-names-for-setting-and-getting-properties-on-Animators,-Shaders-and-Materials) is the description of the inspection I'm getting.

As you said yourself, the Unity HDRP code seems somewhat hacky, so that's probably not the benchmark we should take. But I also don't understand your readability arguments. With a cached prop, it looks like that:

```cs
unityMaterial.SetInt(NormalMapSpace, 0);
```

Non-cached, it looks like that:

```cs
unityMaterial.SetInt("_NormalMapSpace", 0);
```

I would actually prefer the former. Of course, the cached version adds a whole block of property definitions, but they can easily be hidden into a `#region`.

I can refactor this quickly if you wish.

VisualPinball.Unity/VisualPinball.Unity/Import/VpxImporter.cs
freezy: You've marked as resolved, but it's still there?

freezy: Pushed a fix.

PR #142, opened 2020-08-05
Importwizard

Roland09:
Created an import wizard for development mode in order to facilitate the testing of multiple tables. A new popup menu item got added:

popup

It allows you to set a custom table name and it remembers the last loaded table, so you don't have to navigate there anymore. Optionally a favorites list might be handy in the future.

editor

Preferences are saved as Unity EditorPrefs.


freezy:
Thanks!

I'll review when we get in #140, since this one seems to based on it.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizardSettings.cs
```diff
@@ -0,0 +1,41 @@
...
+
+		public static string VpxPath
+		{
+			get => EditorPrefs.GetString("VpxPath", "");
+			set => EditorPrefs.SetString("VpxPath", value);
+		}
+
+		public static string TableName
+		{
+			get => EditorPrefs.GetString("TableName", "");
```
freezy, line 25: Can we set the default to `%TABLENAME%` here so it defaults to the standard import behavior?

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizard.cs
```diff
@@ -0,0 +1,117 @@
...
+	{
+		Window = (VpxImportWizard)GetWindow(typeof(VpxImportWizard));
+		Window.autoRepaintOnSceneChange = true;
+		Window.minSize = new Vector2(800, 300);
+		Window.titleContent = new GUIContent("Visual Pinball Import Wizard"); // here we could attach an icon
+		Window.Show();
+	}
+	#endregion Menu
+
+	public void OnGUI()
```
freezy, line 24: I would prefer a little note letting the user know that it's a dev tool and should only be used when "knowing what doing". Otherwise we'll have complaints about table scripts not working when people use the wizard to import and re-export.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportEngine.cs
```diff
@@ -0,0 +1,50 @@
...
+using NLog;
+using UnityEditor;
+using UnityEngine;
+using VisualPinball.Unity.Import;
+using VisualPinball.Unity.Import.Job;
+using Logger = NLog.Logger;
+
+namespace VisualPinball.Unity.Editor.Import
+{
+	public static class VpxImportEngine
```
freezy, line 11: Can we maybe merge `VpxImportEngine` and `VpxMenuImporter` and just call it `VpxImporter`? `VpxMenuImporter.ImportVpxEditorMemory` is obsolete now anyway, and otherwise there's much code in there.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizard.cs
```diff
@@ -0,0 +1,117 @@
...
+			//GUI.backgroundColor = Color.red;
+			if (GUILayout.Button("Import", GUILayout.Width(100), GUILayout.Height(30)))
+			{
+				if (File.Exists(VpxImportWizardSettings.VpxPath))
+				{
+					VpxImportEngine.Import(VpxImportWizardSettings.VpxPath, null, VpxImportWizardSettings.ApplyPatch, VpxImportWizardSettings.TableName);
+				}
+				else
+				{
+					Debug.LogError(String.Format("VPX file not found: {0}", VpxImportWizardSettings.VpxPath));
```
freezy, line 109: I know it's easier but we have a logger that outputs to `Debug.` when run within Unity. 

```cs
private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
```
Then you can just do:

```cs
Logger.Error("VPX file not found: {0}", VpxImportWizardSettings.VpxPath);
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizardSettings.cs
Roland09: The default (blank) reverts to this, but it's for sure better more verbose if %TABLENAME% is used. I only added it in case someone wants to have that table name in addition to some text like "Pinball [Table1]" for "Pinball [%TABLENAME%]"

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizard.cs
Roland09: Sure, I'll add it.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportEngine.cs
Roland09: In my opinion the importer should be a dedicated class, it should have got nothing to do with the menu. That way one can use it everywhere. This was my use case, i. e. have 1 class that does the import, it's called in the import menu and from the import wizard window and wherever anyone prefers, e. g. from within a script that imports 5 tables just to test out how the importer works.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizard.cs
Roland09: Oh, just habit, my bad. I'll change it.

Roland09:
> 
> 
> Cheers! :)
> 
> I'm still a little hesitant with the table name because it's asking for trouble, but I can understand your use case. So for now it's okay to get this in, and the patch toggle is actually useful.
> 
> Just a few minor changes from my side, let me know if you prefer me doing them.
> 
> Thanks! 👍

I made the requested changes and committed them.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportEngine.cs
freezy: Yeah, I don't like it how it's right now, we have `VpxImportEngine`, `VpxImportWizard`, `VpxMenuImporter`, `VpxImporter`, some are `MonoBehaviour`, some static classes, it's a bit messy.

It's not your fault though, and I'll merge this and create a refactor issue.

VisualPinball.Unity/VisualPinball.Unity.Editor/Import/VpxImportWizard.cs
freezy: No need for `String.Format` though.

freezy: (Pushed fix)

PR #143, opened 2020-08-06
Image manager

rbxnk:
This includes a refactor of the work done for the material manager, generalizing where possible so we have common code for the "manager" windows that will come in the future (sounds, collections, etc).
Image manager itself is pretty straightforward, but our VPE textures were stored in two places, making things a bit confusing and harder to maintain, so I've also introduced generic storage for table global things, and provided a texture specific implementation that replaced the VPE table's container once serialized.


VisualPinball.Engine/VPT/Table/Table.cs
```diff
@@ -177,11 +175,18 @@ public Material GetMaterial(string name)
...
+			Textures = container;
+		}
+
 		public Texture GetTexture(string name)
 		{
-			return name == null
+			string lowerName = name.ToLower();
+			var tex = name == null
 				? null
-				: Textures.ContainsKey(name.ToLower()) ? Textures[name.ToLower()] : null;
```
freezy, line 184: Can `Textures` *not* contain the name? If not, I prefer a fast fail with a descriptive exception instead of an NPE somewhere later.

freezy: Sorry, misread the diff, that's actually what you're doing now.

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
```diff
@@ -0,0 +1,288 @@
...
+	/// </summary>
+	/// <typeparam name="T">class of type IManagerListData that represents the data being edited</typeparam>
+	public abstract class ManagerWindow<T> : EditorWindow where T: class, IManagerListData
+	{
+		protected virtual string DataTypeName => "";
+
+		protected virtual void OnButtonBarGUI() { }
+		protected virtual void OnDataDetailGUI() { }
+		protected virtual void RenameExistingItem(T data, string desiredName) { }
+		protected virtual void CollectData(List<T> data) { }
```
freezy, line 23: Would be more comfortable to understand if it was `protected virtual List<T> CollectData(List<T> data) => new List<T>()`, no?

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
```diff
@@ -0,0 +1,288 @@
...
+				_listView?.SetData(null);
+			}
+
+			if (!string.IsNullOrEmpty(_forceSelectItemWithName)) {
+				_listView.SelectItemWithName(_forceSelectItemWithName);
+				_forceSelectItemWithName = null;
+			}
+
+			EditorGUILayout.BeginHorizontal();
+			if (IsImplemented("AddNewData") && GUILayout.Button("Add", GUILayout.ExpandWidth(false))) {
```
freezy, line 81: I learned about `[GUI.enabled](https://docs.unity3d.com/ScriptReference/GUI-enabled.html)` yesterday - maybe disable the button when there's no implementation?

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
```diff
@@ -0,0 +1,288 @@
...
+			}
+			return acceptedName;
+		}
+
+		// check is a concrete class implements the given method name
+		private bool IsImplemented(string methodName)
+		{
+			var mi = this.GetType().GetMethod(methodName, BindingFlags.Instance | BindingFlags.Public | BindingFlags.NonPublic);
+			return mi != null && mi.GetBaseDefinition().DeclaringType != mi.DeclaringType;
+		}
```
freezy, line 286: That's executed on each editor frame, no?

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableBehavior.cs
```diff
@@ -68,6 +68,8 @@ public class TableBehavior : ItemBehavior<Engine.VPT.Table.Table, TableData>
 		// note: this cache needs to be keyed on the engine material itself so that when its recreated due to property changes the unity material
 		// will cache miss and get recreated as well
 		private readonly Dictionary<PbrMaterial, UnityEngine.Material> _unityMaterials = new Dictionary<PbrMaterial, UnityEngine.Material>();
+		// keep a list of texture names that need recreation, serialized and lazy so when undo happens they'll be considered dirty again
+		[SerializeField] private List<string> _dirtyTextures = new List<string>();
```
freezy, line 72: `HashSet<string>` would be faster, no?

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
rbxnk: Yes absolutely

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
rbxnk: Personally I would interpret a disabled button as something that *could* be used, just not for the selected item.

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
rbxnk: True, but editor frames at least only happen when something changes. Either way a bad precedent, added caching.

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableBehavior.cs
rbxnk: For the access pattern yeah that would make sense, but since it needs to serialize our only collection option is List or array. In practice this isn't going to really matter anyway, the cost of iterating the texture list would be way overshadowed by loading in the real texture data.

freezy: Oh, okay, didn't know that sets weren't serializable.

freezy:
Will play around with it some more, but it looks really sweet already!

image

Just to be sure, the layout manager I've just merged from @Vroonsh (#128) wouldn't be able to make use of your generic manager setup, right? Since there's no details panel anyway.

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ManagerWindow.cs
freezy: You're right, and now I'm reading it again, it doesn't actually show the button if there's no implementation (I thought otherwise).

rbxnk:
> Will play around with it some more, but it looks really sweet already!
> 
> image
> 
> Just to be sure, the layout manager I've just merged from @Vroonsh (#128) wouldn't be able to make use of your generic manager setup, right? Since there's no details panel anyway.

Yeah I don't think its a terribly good fit to share that part of the code, this is really geared towards making a list-style control without foldouts etc. If we have another use case for something similar we could extract the list part away from the detail view.

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ImageManager.cs
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		private void ReplaceImageFromPath(TextureData textureData, string path) {
+			if (_table == null || textureData == null || string.IsNullOrEmpty(path)) { return; }
+
+			byte[] newBytes = null;
+			try {
+				newBytes = File.ReadAllBytes(path);
+			} catch (Exception ex) {
+				Debug.LogException(ex);
```
freezy, line 170: We have our own logger, it would be better to use that one.

```cs
private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
```

```cs
Logger.Error(ex);
```

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ImageManager.cs
```diff
@@ -0,0 +1,232 @@
...
+
+		private void ExportImage()
+		{
+			if (_table == null || _selectedItem == null) { return; }
+
+			var unityTex = _table.GetTexture(_selectedItem.TextureData.Name);
+			if (unityTex != null) {
+				string fileExt = Path.GetExtension(_selectedItem.TextureData.Path).TrimStart('.');
+				if (string.IsNullOrEmpty(fileExt)) {
+					Debug.LogError("Could not determine filetype from path");
```
freezy, line 208: Use NLog logger

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/ImageManager.cs
```diff
@@ -0,0 +1,232 @@
...
+
+		private void UpdateAllImages()
+		{
+			foreach (var t in _table.Textures) {
+				ReplaceImageFromPath(t.Data, t.Data.Path);
+			}
+		}
+
+		private void ReplaceImageFromPath(TextureData textureData, string path) {
+			if (_table == null || textureData == null || string.IsNullOrEmpty(path)) { return; }
```
freezy, line 164: I would also check whether the file exists, because those file paths typically come from other authors. If you start with the default table and refresh images, you'll get 60 error messages.

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableBehavior.cs
```diff
@@ -68,6 +68,8 @@ public class TableBehavior : ItemBehavior<Engine.VPT.Table.Table, TableData>
 		// note: this cache needs to be keyed on the engine material itself so that when its recreated due to property changes the unity material
 		// will cache miss and get recreated as well
 		private readonly Dictionary<PbrMaterial, UnityEngine.Material> _unityMaterials = new Dictionary<PbrMaterial, UnityEngine.Material>();
+		// keep a list of texture names that need recreation, serialized and lazy so when undo happens they'll be considered dirty again
```
freezy, line 71: Not extremely relevant, but when you document a field, I would prefer the triple slashes. They better separate the member and also show up in the editor hover documentation:

```cs
/// <summary>
/// Keeps a list of texture names that need recreation, serialized and
/// lazy so when undo happens they'll be considered dirty again.
/// </summary>
[SerializeField] 
private List<string> _dirtyTextures = new List<string>();
```

rbxnk:
> * When replacing an image, it should filter by HDR/non-HDR. Basically when you're in a non-HDR image, it shouldn't list HDR images in the texture picker, and vice versa. VP will complain if it gets a HDR texture for non-environmental or ball textures. Not sure if this is easily doable though.

It doesn't appear there is a way to filter the picker like that, we can only limit by asset type. Now if it turns out the only use case for hdr/exr in the vpx world is cube maps, we could just limit to picking those and extract the texture data out. Alternatively we could just skip the image replace on after picking if we don't like the extension.
Either way I'd say we can just revisit once we sort out the HDR importing issues and get them rendering correctly.

> * If I drag and drop a new image into my Asset folder, I can select it under replace image. However, when I edit the image and try to pick it again, it shows me the old version in the texture picker (it also shows the old version in image manager, but that's fine / the goal). I was under the impression that Unity automatically refreshes assets as they change? If I "update all" in the image manager, the picker also shows the edited version.

Hmm it sounds like in that case Unity's file watcher just failed to notice the change. If you right click the file and choose "Reimport" does it update?


> * What's with the dirty textures? You mark a texture as dirty when it's replaced, but the only effect is that you're retrieving it from a different Dictionary. In there you're using a `forceRecreate` boolean, but I don't see any "recreation"? Am I missing something?

I've added some comments there to hopefully make it a bit easier to grok, but its not pulling from a *different* dictionary, it's just skipping the cache in that case. `_unityTextures` is a cache of everything we've converted to a Texture2D so far, so when we're dirty we bypass that and grab the raw data from the table again.


> * What's the benefit of `ITableResourceContainer`? If it's just having a more elegant way of dealing with named resources, that's fine, but I'm wondering if there are performance benefits as well?

The primary motivation for introducing this was that we had multiple collections for the table's texture data. We need a serialized list or array in Unity's world so that we can all the normal Unity stuff to it, tracking undos, persisting to disk, etc. but we also need a plain C# collection so the table reader code doesn't have a hard dependency on Unity. So ITableResourceContainer abstracts out the operations we'd need to do on a collection like this, and provides us a single source of truth/single place to go look for our data. So no, long story short, this isn't about performance so much as making maintenance a little easier on us.

> * Not directly related to this PR, but can we make the sidecar GameObject somehow show up last in the hierarchy? Or hide it completely. It happened a few times that I clicked by accident on it while having a big table open, and it froze the editor for minutes.

Experimented with this a bit more, looks like we get the same benefits by using a serialized object instance, let me know if you see the hitching/stuttering come back.

PR #147, opened 2020-08-17
Add Toolbox to Editor

freezy:
This PR adds a new toolbox editor window which allows creating new game items within the currently open table. It also allows creating a new table from scratch.

You can open it through the *Visual Pinball* menu in the editor. It also comes with a brand new set of icons:

image

## Storage Names

Before this PR, the storage name (the name under which the game item is stored in the `.vpx` file, e.g. "GameItem12"), was simply saved and re-used as-is when exporting the table.

Now, when adding and removing items, that doesn't work anymore. The way VP handles it is to keep one big list of all game items and removing and adding elements to that list. When saving, the storage IDs just go from 0 to n-1 in the list's order.

We'd like to do the same, but we don't have a single list of game items. What we have are Dictionaries for each item type (which are unordered). To solve this, this PR removes exposure from these dictionaries and replaces them with methods for accessing and removing game items. Those methods keep the storage indices in order. They are:

- `Flipper(string name)` - Returns a flipper by name. There's a method for each item type.
- `Add<T>(T item, bool updateStorageIndices = false)` - Adds a new item.
- `Has<T>(string name)` - Checks whether an game item of a given type exists.
- `Remove<T>(string name)` - Removes a game item
- `GetAll<T>()` - Returns a list of all game items of a given type.

Internally, we now don't store the storage name anymore, but a `StoragePrefix` and an index. When adding or removing items, we update the indices as necessary. In order to do that, there is now a new `IItem` interface that you can use if you want to refer to a game item in a non-generic way (for `Item<T>` you always needed to specificy the type).

## Default Items

When creating a new game item, it needs some default values. All the supported game items now contain a static `GetDefault(Table table)` method that instantiates a new item with the same default properties as VP.

Part of the new default values are given through a new constructor in the `ItemData`s, and some are just new default values in the data classes themselves. 

Note that all strings are now initialized to `string.Empty`, since `null` isn't actually supported.

## Import

The `VpxImporter` is now called `VpxConverter`, because that's what it actually does, and now we're also using it when creating new items in the editor, so `Importer` was confusing.

## Physics Changes

- We now assert that all colliders have their game item entity assigned. This resulted in runtime errors before, and there were quite a few missing.
- We now ignore frozen balls on top of each system, which eliminated a few other problems.
- Fixed a nasty bug due to handling -0f and 0f the same.

## Other Changes

- Redundant qualifiers were globally removed.
- Changed package naming to Unity's conventions. We now only have one namespace per project, i.e. `VisualPinball.Engine.Unity` for runtime, `VisualPinball.Unity.Editor` for the editor and `VisualPinball.Unity.Patcher` for the patcher.

## TODOs

We need to properly manage multiple tables per scene. We don't need to be able to *play* multiple tables, but something that marks a table as "active" so we know which table to add items to. Issue coming up.


freezy:
If anyone interested to review: Look at `ToolboxEditor.cs` and the updated `Behavior`s.

freezy:
All editors open:

image

PR #153, opened 2020-08-18
Add more patcher features

### LINKED ISSUE #145: Patcher Refinements
### Patch file detection

Currently the patch file is determined by meta tags. This isn't fine granular enough, e. g. the Tom & Jerry table has the wrong meta tags:

`[MetaMatch(TableName = "Beach Bums (HH Mod - Gottlieb 1986)", AuthorName = "Retro Bash")]`

However the table's name is `TomandJerry` which can in addition to the `MetaMatch` be used to get the proper patch file.

### Access to Primitive data

There needs to be a re-parenting of gameobjects. In order for that to work, one also needs to apply the transform of the Primitive. That data isn't available in the patcher, it should be handed over in addition to the gameobject.

### Patching children

The Tom & Jerry table has a gameobject `Ramp20` which has children `LeftWall`, `RightWall`, etc. The way it is now one has to set NameMatch to `Ramp20` and the code in the patching methoc would be applied to all the children, not just a selected one.

This should be made finer granular, e. g. `Ramp20/LeftWall`. Of course only optionally. `Ramp20` should still select all children.

### Disable patching during Import

I added a parameter `applyPatch` to the importer, but it's unclear where to put that if-code in order to not apply a patch. For one because at certain places the importer uses local methods, for the other it uses static methods.

### Graphics Pipeline

Depending on the graphics pipeline one has to set different shader properties. Should we distinguish by creating a dedicated file for the graphics pipeline? Or a common interface which allows to set e. g. a material to be double sided. In HDRP this might look like this:

```
[NameMatch("Primitive66")] // jerry at plunger
public void SetDoubleSided(GameObject gameObject)
{
	var unityMat = gameObject.GetComponent<Renderer>().sharedMaterial;

	unityMat.EnableKeyword("_DOUBLESIDED_ON");
	unityMat.EnableKeyword("_NORMALMAP_TANGENT_SPACE");

	unityMat.SetInt("_DoubleSidedEnable", 1);
	unityMat.SetInt("_DoubleSidedNormalMode", 1);

	unityMat.SetInt("_CullMode", 0);
	unityMat.SetInt("_CullModeForward", 0);
}
```

Which is way too much code to be duplicated per patch file and hence a common HDRP class with SetDoubleSided(...) might be the better solution.

We should decide in advance, this will have far reaching consequences.

freezy:
The patcher is now more flexible. Basically you can pull in more data when applying the patch, and there is a new matcher that matches by the script table name (`TableNameMatch`) and one by render pipeline (`RenderPipeline`). The [patcher's README](https://github.com/freezy/VisualPinball.Engine/tree/patcher/enhancements/VisualPinball.Unity/VisualPinball.Unity.Patcher#unity-patching-system) is updated, so you can check for details there, but the most relevant part is:

> You can pass different types in any order to the method. Supported types are:
> 
> - `UnityEngine.GameObject` - Unity game object that was created for that game item.
> - `VisualPinball.Engine.VPT.Table.Table` - The table object for which the game 
>   item was created for
> - Any game item type, e.g. `VisualPinball.Engine.VPT.Flipper.Flipper` that 
>   extends `IItem`. If your matched game item is not of the type you've provided
>   in the method signature, the patch is skipped and a warning is printed.
> - `VisualPinball.Engine.Game.IRenderable` if you don't care about the item type
>   but still want to access something from the item.

Additionally, patches are now all applied *after* all game items are created, which was necessary if a game item needs re-parenting (i.e. moved below another game item, which needs to exist).

This PR closes #145.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Toolbox/ToolboxEditor.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Toolbox/ToolboxEditor.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using UnityEditor;
 using UnityEngine;
 using VisualPinball.Engine.Game;
@@ -193,7 +194,8 @@ namespace VisualPinball.Unity.Editor
 		{
 			var tb = TableAuthoring;
 			var rog = renderable.GetRenderObjects(table, Origin.Original, false);
-			return VpxConverter.ConvertRenderObjects(renderable, rog, GetOrCreateParent(tb, rog), tb);
+			VpxConverter.ConvertRenderObjects(renderable, rog, GetOrCreateParent(tb, rog), tb, out var obj);
+			return obj;
 		}
 
 		private static GameObject GetOrCreateParent(Component tb, RenderObjectGroup rog)
--- a/VisualPinball.Unity/VisualPinball.Unity.Patcher/Matcher/Item/ItemMatchAttribute.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Patcher/Matcher/Item/ItemMatchAttribute.cs
@@ -7,6 +7,11 @@ namespace VisualPinball.Unity.Patcher
 	[AttributeUsage(AttributeTargets.Method, AllowMultiple = true)]
 	public abstract class ItemMatchAttribute : Attribute
 	{
+		/// <summary>
+		/// If set, pass the game object with this name as reference to the patcher.
+		/// </summary>
+		public string Ref;
+
 		public abstract bool Matches(Engine.VPT.Table.Table table, IRenderable item, RenderObject ro, GameObject obj);
 	}
 }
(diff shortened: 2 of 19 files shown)


freezy:
Okay, I've now added this to JP:

```cs
[NameMatch("LFLogo", Ref="Flippers/LeftFlipper")]
[NameMatch("RFLogo", Ref="Flippers/RightFlipper")]
[NameMatch("RFLogo1", Ref="Flippers/UpperRightFlipper")]
public void ReparentFlippers(GameObject gameObject, ref GameObject parent)
{
	var rot = gameObject.transform.rotation;
	var pos = gameObject.transform.position;

	// re-parent the child
	gameObject.transform.SetParent(parent.transform, false);

	gameObject.transform.rotation = rot;
	gameObject.transform.position = pos;
}
```

That okay?

Roland09:
Looks great to me. Gonna introduce some class with common patching functions.

freezy:
Thanks. Can you have a quick look through the code? I'll merge when you approve.

Roland09:
Looks great, thanks for your efforts.

freezy:
Thanks! There's an "approve" button btw ;)

Merging!

PR #154, opened 2020-08-19
Update Dependencies. Update "Vendor" folder to "Plugins" to match Unity conventions.

jsm174:
This PR has the following updates:

- Update `NLog` to `4.7.3`
- Update `JeremyAnsel.Media.WavefrontObj` to `2.0.19`
- Update `Vendor` folder to `Plugins` to match Unity conventions
- Update `Plugins` copy of `OpenMcdf` to `2.2.1.6`
- Update `NetMiniZ` to `1.1.1`
- Removes building any "managed" DLLs as x86 or x64 and uses AnyCPU instead. 
- Update the github workflow to list build folder contents as a separate step
- Update the github workflow `visualpinball-engine-win-x86` job to build with `PlatformTarget=x86`, so the correct native libraries are placed in build folder
- Update the github workflow `visualpinball-engine-win-x86` job to launch nunit console runner as x86 process
- Clean up solution files for consistency


freezy (approved):
Sweet, thanks a lot!

PR #157, opened 2020-08-20
Layer Editor Improvements

### LINKED ISSUE #150: Layer Manager Improvements
There are a few things in the layer manager I'd like to be improved:

- [x] When opened, it should always expand the layers.
- [x] Clicking on the plus icon, should prompt for a layer name. When added, the new layer should be selected.
- [x] Same when right-clicking on a game item and selecting *assign to / new layer*: Prompt name and select created layer.
- [x] Dragging a game object from the Hierarchy into a layer doesn't work.
- [x] Remove the *Select only game items to enable layer assignment* menu when right-clicking on a layer, I find it confusing
- [x] Only show *new layer* when right-clicking on the top table node
- [x] Only show *delete layer* when right-clicking on a layer
- [x] When adding a new item with the toolbox, the item should be added to the currently selected layer (or, if an item is selected, the layer of the item). When multiple layers are selected, the first layer should be used.
- [x] Selection management: Double-click to select in the Hierarchy is fine, but what about multiple elements? Let's say an author keeps all GI lights in a layer and want to update their properties simultaneously. Multi-selecting them in the layer and having them multi-selected in the Hierarchy would be nice. Maybe with a right-click menu?
- [x] Selection vol2: What if an item is selected in the hierarchy? How to find it rapidly in the layer manager? Would an auto-select in layer manager when selection changes in the hierarchy be possible?

Vroonsh:
Remark : Had to create a ItemCreated event in ToolBoxEditor because EditorWindow.OnHierarchyChange() was not providing any context.

Should close #150

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/ItemInspector.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/ItemInspector.cs
@@ -15,6 +15,8 @@ namespace VisualPinball.Unity.Editor
 		protected string[] _allMaterials = new string[0];
 		protected string[] _allTextures = new string[0];
 
+		public static event Action<IIdentifiableItemAuthoring, string, string> ItemRenamed;
+
 		protected virtual void OnEnable()
 		{
 			_table = (target as MonoBehaviour)?.gameObject.GetComponentInParent<TableAuthoring>();
@@ -50,7 +52,9 @@ namespace VisualPinball.Unity.Editor
 		{
 			if (target is MonoBehaviour bh && target is IIdentifiableItemAuthoring item && bh != null) {
 				if (item.Name != bh.gameObject.name) {
+					var oldName = item.Name;
 					item.Name = bh.gameObject.name;
+					ItemRenamed?.Invoke(item, oldName, bh.gameObject.name);
 				}
 			}
 		}
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Layers/LayerEditor.cs
@@ -1,3 +1,4 @@
+using System.Collections.Generic;
 using System.Linq;
 using UnityEditor;
 using UnityEditor.IMGUI.Controls;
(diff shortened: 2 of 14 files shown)


freezy:
Great stuff. Things I've noticed:

- When dragging an item from the hierarchy to an expanded layer, the layer collapses when dropped, hiding the item.
- I don't see a way to multi-select items in the hierarchy through the layer manager. You've commented that double-clicking works, but that's only for single items, or am I missing something?

Vroonsh:
> * When dragging an item from the hierarchy to an expanded layer, the layer collapses when dropped, hiding the item.
>
Yeah saw that also, treeview weirdness, i just ask for the layer item to be expanded (this autoexpand feature is totally automatic, maybe it's messing with the TreeViewState), i'll check that

> * I don't see a way to multi-select items in the hierarchy through the layer manager. You've commented that double-clicking >works, but that's only for single items, or am I missing something?
>
When double clicking a layer item, it's selecting all items into it, same while double clicking the Table Item.
There is no multi double click behavior, or maybe some kind of Ctrl+double click but it would be weird, i'll check that also

freezy:
The idea was that if you have multiple selected items (or even layers) in the layer manager, you'd have a right-click option "select in hierarchy" where it selects all of them so you can edit common properties in the inspector.

Vroonsh:
Updated the Double click behavior with Ctrl modifier support
For the other issue that's weird, debugging the thing the TreeViewState is correct (ExpandedIds -1(Root), 0(Table) and layernum) but the layer is collapsed.

freezy:
Cool. You're sure CTRL is the best way to deal with it? Isn't right-click -> menu more intuitive?

Vroonsh:
> Cool. You're sure CTRL is the best way to deal with it? Isn't right-click -> menu more intuitive?

Because Ctrl modifier was already related to multiselection for single click i found that quite intuitive.
But i can add a context menu entry afterward if you wish.

freezy:
Okay, having both ways would be the best solution IMO.

freezy:
Thanks a lot (and sorry again for being a PITA 😄)!

Will re-test and merge tonight!

Vroonsh:
> Thanks a lot (and sorry again for being a PITA 😄)!
> 
> Will re-test and merge tonight!

Haha ! no worries, VPE has to be as good as possible ;)

Vroonsh:
Added a Framing helper to Unity.Editor.Utils
Wanted to use SceneView.FrameLastActiveSceneView but it has some weird behavior when selecting a layer or the table.

Vroonsh:
Added an ItemRenamed event to the ItemInspector, catched by the LayerTreeView to reload.
This event will also be usefull e.g. for Collections Editor (collections are based on Item names apparently)
I've also a local commit on layers branch for a RenameValidation event which could be usefull for validating our gameObjects renaming by our TableAuthoring.
It's already renaming the gameObject with its old name if any validator disagree with the new name.
Dunno if you want me to add this one to the PR ?

freezy:
Another PR please, merging now!

PR #167, opened 2020-08-27
Mesh animation frames into Unity blendshape

ecurtz:
This code reads the VPX vertex and normal animation frames and builds a Unity blendshape from the data. This also involves adding a SkinnedMeshRenderer rather than a normal MeshRenderer component on objects with vertex animation.


ecurtz:
Removed Unity.Vector3 references from VPE base files.

freezy:
Awesome, thanks a lot! ❤️ 

Any chance of adding a few unit tests? This concerns the data layer which is pretty well tested. The most important one would be:

1. Create a simple animation in Blender
2. Export it
3. Import it into an empty VP table in VPX
4. Export it in VPX as `.obj`

In the unit test, load the table, and assert that the imported mesh animation is the same as the exported `.obj`. There are helper which can load `.obj` files, but you probably have to create a new method for comparing animation frames.

Let me know if that's okay, otherwise it would be great to have at least the animation from Blender.

freezy (approved):
Thanks @ecurtz and welcome to the team! :)

PR #169, opened 2020-08-30
drag points : Control points fixes

Vroonsh:
- Fixed missmatch result for ClosestPointToPolyline on all straight polylines
- Fixed wrong ratio calculation on control point add (for ramps height calculation)


freezy (approved):
Magnifique, merci beaucoup! 👍

PR #171, opened 2020-08-30
Port last collider: Hit3DPoly

### LINKED ISSUE #123: Handle dynamically sized colliders (Poly3DCollider)
We currently can't instantiate blob pointers with an arbitrary size. I've created a thread [here](https://forum.unity.com/threads/blobbuilder-allow-explicit-length-when-allocating-blobptr.875917/). Until this is solved, Poly3DCollider don't work.

freezy:
This implements the `Hit3DPoly` collider. It closes #123. Just to be clear what `Hit3DPoly` does: It's a 2D-polygon in the XY-plane and it collides with its surface, i.e. against the Z axis. It's only used in three cases:

- Surface: the ceiling and optionally the ground (depending on `IsBottomSolid`) of the surface
- Trigger: the surface under the trigger that kicks off the events.
- Playfield: the glass surface of the playfield

They way this is solved now is that instead of one dynamically sized polygon, we triangulate it and split it into triangles, edges and points. Like that we can keep the collider size static. 

The only problem is now the event check for the trigger, since one collider now doesn't know whether the ball has entered into the *combination* of all triangles. We could solve this by marking the triangle line segments as inner and outer. Then, only when an *outer* line is crossed, we update the `BallInsideOfBufferElement`s.

For the triangulation we're using Nick Gravelyn's [Triangulator](https://github.com/nickgravelyn/Triangulator). For generating the colliders out of the triangulated mesh, we're re-using the primitive's hit generator.

## TODO
- [ ] Fix trigger event
- [x] Clean up obsolete `Poly3DCollider` code

### DIFF
--- a/VisualPinball.Engine/VPT/Table/Table.cs
+++ b/VisualPinball.Engine/VPT/Table/Table.cs
@@ -16,7 +16,7 @@ namespace VisualPinball.Engine.VPT.Table
 	/// A table contains all the playfield elements, as well as a set of
 	/// global data.
 	/// </summary>
-	public class Table : Item<TableData>, IRenderable
+	public class Table : Item<TableData>, IRenderable, IHittable
 	{
 		public CustomInfoTags CustomInfoTags { get; set; }
 		public int FileVersion { get; set; }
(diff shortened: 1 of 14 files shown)


VisualPinball.Engine/VPT/Table/Table.cs
```diff
@@ -251,6 +254,10 @@ public class Table : Item<TableData>, IRenderable
 
 		#endregion
 
+		public void Init(Table table)
```
Vroonsh, line 257: Is it needed ?

Vroonsh (approved):

VisualPinball.Engine/VPT/Table/Table.cs
freezy: It's part of `IPlayable`, so yes. Otherwise there would be a lot of refactoring for not much benefit.

freezy:
Trigger event will be handled in #176.

PR #177, opened 2020-09-02
Clean up legacy physics

freezy:
This removes all the physics code that was replaced by the DOTS implementation.


freezy:
Note the removed code will stay on [legacy/generic-physics](https://github.com/freezy/VisualPinball.Engine/tree/legacy/generic-physics).

PR #184, opened 2020-09-06
Patcher/utilities

Roland09:
Trying to keep patching code at a minimum and centralized by providing a common utility class. Current implementation is HDRP only, URP and Built-In will come later.


freezy (approved):
Looks good, thanks! 👍

PR #186, opened 2020-09-07
Mult table editing

rbxnk:
This PR adds support for multiple tables in a scene. 
* Allow moving root object as a normal unity transform
* Add "lock" icon (like inspectors) to our manager windows
* Does not add support for multi tables in runtime/player (went down this rabbit hole a bit, we'd need to totally restructure the physics systems to not auto create)


VisualPinball.Unity/VisualPinball.Unity.Editor/Toolbox/ToolboxEditor.cs
```diff
@@ -205,21 +201,20 @@ private static bool CreateButton(string label, Texture icon, float iconSize, GUI
 			);
 		}
 
-		private static void CreateItem<TItem>(Func<Table, TItem> create, string actionName) where TItem : IItem
+		private static void CreateItem<TItem>(TableAuthoring tableAuth, Func<Table, TItem> create, string actionName) where TItem : IItem
```
freezy, line 204: It's only `static` because previously wasn't using any members. Now I would suggest making it non-static and using `_table` directly instead of passing it through every call.

VisualPinball.Unity/VisualPinball.Unity.Editor/Toolbox/ToolboxEditor.cs
```diff
@@ -205,21 +201,20 @@ private static bool CreateButton(string label, Texture icon, float iconSize, GUI
...
 			var item = create(table);
 			table.Add(item, true);
-			Selection.activeGameObject = CreateRenderable(table, item as IRenderable);
+			Selection.activeGameObject = CreateRenderable(tableAuth, item as IRenderable);
 			ItemCreated?.Invoke(Selection.activeGameObject);
 			Undo.RegisterCreatedObjectUndo(Selection.activeGameObject, actionName);
 		}
 
-		private static GameObject CreateRenderable(Table table, IRenderable renderable)
+		private static GameObject CreateRenderable(TableAuthoring tableAuth, IRenderable renderable)
```
freezy, line 214: Same here, make it non-static and use `_table` directly.

freezy (commented):
Awesome PR, thanks a lot!

Just two minor suggestions above, and haven't tested it yet.

Also now we have a documentation site, I will add some documentation and let you review it.

Cheers! 👍

rbxnk:
Yup looks good, my first thought was to make a comment about how copying things between the heirarchies wouldn't work yet, but you were way ahead of me :-)

freezy (approved):

PR #192, opened 2020-09-22
biffdata : fixed unserialized StoragePrefix 

Vroonsh:
- leading to StorageName duplication for GameItems/Sounds/Collections/Images at scene reload


freezy (approved):
Cheers. I think those bugs might have come from me just blindly applying IDE suggestions. It if it was, apologies and I'll be more careful in the future (and see how I can adapt them).

Issue #106, opened 2020-07-07
Indicate gate direction in editor


freezy:
VPX shows in which direction a gate is oriented.

One way:

![image](https://user-images.githubusercontent.com/70426/86847596-db0a1380-c0ac-11ea-8190-4b11d36b7286.png)

Two way:

![image](https://user-images.githubusercontent.com/70426/86847622-e8bf9900-c0ac-11ea-911f-06ae56cba025.png)

This would be useful to see in the Unity's Scene View as well.

PR #193, opened 2020-09-22
gates : draw direction gizmo on selected gate

Vroonsh:
should close https://github.com/freezy/VisualPinball.Engine/issues/106

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
@@ -47,7 +47,7 @@ namespace VisualPinball.Unity.Editor
 				var transform = (target as MonoBehaviour).transform;
 				if (transform != null && transform.parent != null) {
 					position = transform.parent.TransformPoint(position);
-					var axis = transform.TransformDirection(Vector3.up);
+					var axis = transform.TransformDirection(-Vector3.up); //Local direction of the gate gameObject is -up
 					var worldScale = 0.5f * VpxConverter.GlobalScale;
 					var scale = _gate.Item.Data.Length * worldScale;
 					Handles.color = Color.white;
@@ -60,9 +60,9 @@ namespace VisualPinball.Unity.Editor
 					Handles.color = Color.white;
 					var ratioscale = 2.0f;
 					var arrowscale = (worldScale * (100.0f - ratioscale)) + Mathf.PingPong(Time.realtimeSinceStartup * worldScale * ratioscale * 2.0f, worldScale * ratioscale);
-					Handles.ArrowHandleCap(-1, position, Quaternion.LookRotation(-axis), arrowscale, EventType.Repaint);
+					Handles.ArrowHandleCap(-1, position, Quaternion.LookRotation(axis), arrowscale, EventType.Repaint);
 					if (_gate.Item.Data.TwoWay) {
-						Handles.ArrowHandleCap(-1, position, Quaternion.LookRotation(axis), arrowscale, EventType.Repaint);
+						Handles.ArrowHandleCap(-1, position, Quaternion.LookRotation(-axis), arrowscale, EventType.Repaint);
 					}
 				}
 				HandleUtility.Repaint();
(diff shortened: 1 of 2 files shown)


Vroonsh:
Oops sorry, i've the commit for reload issues from my master branch also

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
```diff
@@ -37,6 +38,32 @@ protected override void OnEnable()
...
 
+		protected virtual void OnSceneGUI()
+		{
+			if (target is IEditableItemAuthoring editable) {
+				var position = editable.GetEditorPosition();
+				var transform = (target as MonoBehaviour).transform;
+				if (transform != null && transform.parent != null) {
+					position = transform.parent.TransformPoint(position);
+					var axis = transform.TransformDirection(Vector3.up);
+					var scale = _gate.Item.Data.Length * 0.0005f;
```
freezy, line 49: That's world scale, right? If so, I'd use the global world scale constant we have (`VpxConverter.GlobalScale`).

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
```diff
@@ -78,5 +105,14 @@ public override void OnInspectorGUI()
 
 			base.OnInspectorGUI();
 		}
+
+		protected override void FinishEdit(string label, bool dirtyMesh = true)
+		{
+			if (label == "Two Way") {
```
freezy, line 111: Not optimal since we might rename that label and it will break the repaint, but not sure about an alternative.

Vroonsh: I could put a static readonly and use it on both places

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
Vroonsh: didn't know about this one i'll check that

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/GateInspector.cs
freezy: Or a third optional parameter with an enum representing the field? It would only have two values (`Any`, `IsTwoWay`) for now. You can pass it to `ItemInspector.ItemDataField()` and receive it back in `FinishEdit()`.

Vroonsh: A lot of refactor if i change the FinishEdit prototype, maybe in another issue about fields edition ?

freezy: An optional argument would add two lines of code (one to `ItemDataField(string label, ref bool field, bool dirtyMesh = true)`, and one to `FinishEdit(string label, bool dirtyMesh = true)`), no?

freezy: Plus the enum definition of course.

freezy (approved):
Great PR, thanks a lot! :)

freezy:
So as you mentioned the animation seems to be a drag. This is my CPU without and with a gate selected and no other interactions:



For me that's significant enough to ditch the animation. It's a pity because it looks nice, but performance is more important, and it adds no additional value.

freezy:
Hmm why did you close the PR?

Vroonsh:
Just wanted to remove the reload fix commit, will redo it

Vroonsh:
About performances, the animation only reveal the SceneView.Repaint heavy performance hit.
I tried with the default HDRP sample scene and while rotating the camera around the scene, CPU usage goes up to 45%.
So no real way to improve on that except providing extra edition settings (including quality settings/switches not already present in Unity).

PR #195, opened 2020-09-23
dragpoints : performance improve

Vroonsh:
- redraw handles only when curvetraveller has moved  and mouse is over sceneview


freezy (approved):
Nice!

PR #197, opened 2020-09-23
Tidy up physics code

freezy:
Unlike the branch name suggests, this doesn't increase performance, but it refactors the physics code so everything can be run on threads.

I've tried exactly that obviously, and it doesn't help; scheduling some system takes about 0.05ms while running the system on the main thread takes 0.002ms. So it took a huge performance hit:

image

blue (left): `.Run()`
red (right): `.Schedule[Parallel]()`

Some systems might be able to leverage multi-threaded execution, and I'll take some more time, but the tests I've run with 8 balls falling onto a few table elements run at a median 3.95ms per frame, which seems pretty good already.

image


xantari:
Which profiler are you using in the screenshots?

freezy:
[Unity's Profile Analyzer](https://blogs.unity3d.com/2019/05/13/introducing-the-profile-analyzer/)

Issue #187, opened 2020-09-09
Don't show sound source if sound manager is invisible


freezy:
The sound position animation is awesome, but it should only be shown when the sound manager is visible:

![image](https://user-images.githubusercontent.com/70426/92647014-86bb2400-f2e7-11ea-9135-86c67be6b24b.png)

Issue #163, opened 2020-08-26
Handle Sounds in Editor


Vroonsh:
Sounds are red & written but we cannot create new ones or edit existing.

![image](https://user-images.githubusercontent.com/13945459/91357668-e8f92c80-e7f1-11ea-8d01-1c610c9b7be3.png)

![image](https://user-images.githubusercontent.com/13945459/91357697-f3b3c180-e7f1-11ea-8b64-a223d827a804.png)

PR #200, opened 2020-09-27
sounds : SoundManager upgrades

Vroonsh:
- fixed sound position display in SceneView when manager unfocus, close https://github.com/freezy/VisualPinball.Engine/issues/187

- add a display all sounds toggle, selected sound still highlighted
- add an auto framing feature, automatically switch to top view & frame on table when focusing the soundmanager
- change sound position sphere radius regarding SoundData.Volume

image

image

Should close https://github.com/freezy/VisualPinball.Engine/issues/163

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/SoundManager.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/SoundManager.cs
@@ -20,6 +20,7 @@ using UnityEngine;
 using UnityEditor;
 using VisualPinball.Engine.VPT.Sound;
 using VisualPinball.Engine.VPT;
+using VisualPinball.Unity.Editor.Utils;
 
 namespace VisualPinball.Unity.Editor
 {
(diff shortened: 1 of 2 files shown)


freezy (changes requested):
Hey, sorry for the delay. I like the framing feature, good stuff! 👍 

However I checked my CPU usage, and even on the desktop it's pretty bad:



How about showing a speaker icon at the sound location instead of the animation? I know I previously agreed to the sound source animation, but redrawing the scene constantly is quite a load, and it will be choppy on low-end systems.

Vroonsh:
> Hey, sorry for the delay. I like the framing feature, good stuff! 👍
> 
> However I checked my CPU usage, and even on the desktop it's pretty bad:
> 
> 
> 
> How about showing a speaker icon at the sound location instead of the animation? I know I previously agreed to the sound source animation, but redrawing the scene constantly is quite a load, and it will be choppy on low-end systems.

Yup will do, still more than 30% CPU usage on scene view on a i7 8700 amazing !
Pretty sure you'll have the same result with an empty scene and moving the camera as i tried before :(

freezy (approved):
Awesome, like the circle! Cheers!

PR #201, opened 2020-10-04
Better name synchronization between Game Object and ItemData

ecurtz:
More reliable comparison of names by using OnPreInspectorGUI() rather than EditorApplication.hierarchyChanged(), which was not always called for inspector changes.


ecurtz:
Obviously there's something weird in the fact that there are 14 commits involved in my 20 line change. I honestly have no idea what it means that it wants to commit all those unmodified files. If I can figure out what git wants I can fix it or of course somebody who knows what they're doing can make the changes to ItemInspector.cs and do a new pull request.

freezy:
Cheers!

It looks like you didn't pull the latest changes from my repo when you were working on the fix. I'll squash the commits to your latest.

freezy:
Merged in #202.

PR #202, opened 2020-10-04
Better name synchronization between Game Object and ItemData 

freezy:
Same as #201 without the git clutter.


ecurtz (approved):
Looks good, thanks for doing the cleanup.

PR #203, opened 2020-10-07
Delete via multiselection

Roland09:
* multi-selection enabled
* getter for all selected rows
* deletion of all selected rows; single row deletion shows name in confirmation popup, multi-selection shows number of selected items in confirmation popup


freezy (approved):
Looking good, thanks! @rbxnk can you have a look as well please?

rbxnk (approved):

PR #214, opened 2020-10-14
Depth Prepass example for HDRP

Roland09:
* patcher for Mississippi table. required depth prepass enabled or else the images wouldn't be shown from a distance


freezy (approved):
Thanks! :)

PR #215, opened 2020-10-15
Layer for gameobjects

Roland09:
* dedicated class for layers, might differ in the future
* moved layer id from 8 to 15 because of postprocessing layer conflict in hdrp

### DIFF
--- a/VisualPinball.Engine/Common/Constants.cs
+++ b/VisualPinball.Engine/Common/Constants.cs
@@ -28,6 +28,17 @@ namespace VisualPinball.Engine.Common
 		public const float Gravity = 1.81751f;
 	}
 
+	/// <summary>
+	/// Layer of the table items.
+	/// Currently only one layer, but might be more in the future.
+	/// </summary>
+	public static class Layer
+	{
+		public const int ChildObjects = 15;
+	}
+	
+
+
 	public static class PhysicsConstants
 	{
 		public const float HitShapeDetailLevel = 7.0f;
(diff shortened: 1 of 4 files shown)


VisualPinball.Engine/Common/Constants.cs
```diff
@@ -28,6 +28,17 @@ public static class Constants
 		public const float Gravity = 1.81751f;
 	}
 
+	/// <summary>
+	/// Layer of the table items.
+	/// Currently only one layer, but might be more in the future.
+	/// </summary>
+	public static class Layer
```
freezy, line 35: Can you name it `LayerConstants`?

freezy (commented):
Thanks, but I'm probably not going to merge this. The reason is because I have a huge refactoring branch where the blocks of your diff don't exist anymore, and I prefer to get this into my branch manually.

VisualPinball.Engine/Common/Constants.cs
Roland09: Ok. Main point of this is that you change this:

public const int ChildObjectsLayer = 8;

to anything larger than 8 because 8 is used as PostProcessing Layer in HDRP by default.

Roland09:
> 
> 
> Thanks, but I'm probably not going to merge this. The reason is because I have a huge refactoring branch where the blocks of your diff don't exist anymore, and I prefer to get this into my branch manually.

Ok. Main point of this is that you change this:

`public const int ChildObjectsLayer = 8;`

to anything larger than 8 because 8 is used as PostProcessing Layer in HDRP by default.

Roland09:
Will be changed later, not relevant right now.

PR #219, opened 2020-10-16
Implement Coil Manager

### LINKED ISSUE #212: Add Coil Manager
Similar to the switch manager, we would like to assign coils to the game logic engine visually. It's pretty much the same feature, but where switches were "sources", coils are now "destinations", so the naming changes little. In terms of logic, it's the same.

coil_manager

We should include the additional sources from #211, which are additional destinations now, so you'd have another interface:

```cs
public interface ICoilDevice
{
    string Name { get; }
    IEnumerable<GamelogicEngineCoil> Coil;
}
```

Since there are no other destinations but the playfield and ICoilDevices, we can do without all the input and constant stuff. So a data entry class could look something like that (BIFF attributes removed):

```cs
public class MappingInputEntryData : BiffData
{
    public string Id = string.Empty;
    public string Description = string.Empty;
    public int Destination = CoilDestination.Playfield;
    public string PlayfieldItem = string.Empty;
    public string DeviceDestination = string.Empty;
    public string DeviceItem = string.Empty;
    public int Type = SwitchType.OnOff;
    public int Pulse = 10;
}
```

Note that there will be another issue about directly linking switches to coils.

jsm174:
This PR fixes #212.

Updates include:

- Added Coil Manager
- Simplified `MappingConfig` to `Mappings`
- Removed having to load `Mappings` from sidecar as a dictionary using "Switches" key
- Refactored all references of `Switchable` to `Switch`. (Since `Coilable` and `Wireable` do not really fit, it is for consistency)
- Updated `Mappings` test fixture with sample coils
- Added `Device` and `DeviceItem` fields to Biffs in preparation for devices such as troughs.
- Added `IGameEngineAuthoring` interface
- Updated `Player` to use `IGamelogicEngine` instead of `DefaultGamelogicEngine`
- Added `Coin Door Enter (WPC)` (0 key) action to `InputManager` 

As requested, this PR has been rebased against `fix/playmode`.

If it is not correct, original branch against master is at: 

https://github.com/jsm174/VisualPinball.Engine/tree/feature/coil_manager


freezy (approved):
Awesome, thanks a lot @jsm174!

Before I merge, I'll:

- [ ] Add kicker config how to kick (it's not a setting currently)
- [ ] Make sure "populate all" works with the default table and update setup guide

freezy:
I think we should be fine now. The screenshots in the documentation aren't up-to-date, but I'll do that when devices are done.

@jsm174 if you could just proof-read my latest changes in `switch-manager.md` and confirm that the coil example of MM is correct, then I'll merge.

jsm174:
> I think we should be fine now. The screenshots in the documentation aren't up-to-date, but I'll do that when devices are done.
> 
> @jsm174 if you could just proof-read my latest changes in `switch-manager.md` and confirm that the coil example of MM is correct, then I'll merge.

Proof-read `switch-manager.md`. All good. 

The screenshot for MM seems 100% correct, however, WPCEmu is reporting different switches than the manual. So I will continue to research.

PR #221, opened 2020-10-16
layers : filter objects in scene view while filtering in LayerEditor

Vroonsh:
- Create new BaseEditorWindow as parent for our editor window classes
- Moved common editor window classes into Unity.Editor.Editors directory
- Add SearchableEditorWindow support to BaseEditorWindow with hidden fields/methods exposure, disabled by default
- Setup LayerEditor to be in the same search group as SceneHierarchyWindow & SceneView
- Only shows Items in LayerEditor treeview when filtering

image


freezy:
Hey @Vroonsh, the coil manager merge too longer than expected, could you rebase and resolve the conflicts, then I'll review! Thanks!

Vroonsh:
> Hey @Vroonsh, the coil manager merge took longer than expected, could you rebase and resolve the conflicts, then I'll review! Thanks!

Sure, will do

freezy (approved):
Cool! This uses reflection though, right? @rbxnk could you have a look as well to check whether this might not break something else?

Cheers!

Vroonsh:
Yup I had to use unexposed methods & fields from the SearchableEditorWindow to synchronize the search field. 

It uses the same searchfied Gui as hierarchy & sceneview windows are using.

PR #228, opened 2020-10-28
Initial trough implementation

ecurtz:
The only changes to actual code were making the TableApi object accessible in Player and ItemApi and modifying ItemInspector for a generic way of linking to other VPE "Authoring" classes. The rest of it is new trough code. It is NOT fully functional at this point, but that's partially because KickerApi is still WIP.


freezy:
Closing in favor of #229.

PR #229, opened 2020-10-30
Add Trough Component

### LINKED ISSUE #211: Switch Manager: Add support for additional sources
We'll have game mechanics like troughs that have multiple switches (and coils). I think the best way to support them without making everything incredibly complicated is to interpret them as "source".

I would imagine having an interface like that:

```cs
public interface ISwitchDevice
{
    string Name { get; }
    IEnumerable<GamelogicEngineSwitch> Switches;
}
```

Our trough component would then implement this interface, and it thus appear in the switch manager as additional source:

switch-manager2

Data-wise, we have two possibilities:

1. Change `Source` from int to string, and introduce "special" values for input system, playfield, constant (maybe prefix with double underscore?)
2. Add an other `DeviceSource` field that is a string, and add another `Source` value (e.g. -1 = use `DeviceSource`)

I would prefer 2. since it's more proper. We also need to store the actual value in a new field, let's say `DeviceItem`.

So for the trough example above, we would have the following data:

- `Id` = "15"
- `Description` = "Trough I"
- `Source` = `SwitchSource.Device`
- `DeviceSource` = "Trough"
- `DeviceItem` = "Switch I"
- `Type` = `SwitchType.OnOff`

### LINKED ISSUE #234: Rubbers: Height, Rot* related issues.
The rubbers appear to have a few issues: 

The collision height value does not update when adjusting the value.  It will update after another value has been updated.  
Height value on geometry appears to have no change on the model.  
Lock tick box does not appear to have any effect.  
The transformation matrix does not appear to be applied.

freezy:
This PR adds support for trough devices and refactors the switch- and coil mapping so it supports multi-switch/coil devices. Closes #211. Supersedes and closes #228. User documentation can be found [here](https://github.com/freezy/VisualPinball.Engine/blob/feature/trough/VisualPinball.Unity/Documentation~/creators-guide/manual/mechanisms/troughs.md).

The UI was solved a bit differently, since I didn't want to overload the source/destination dropdown. Mockup:

Final screen:

So in case of a Device, the *Element* column has now two fields, one for the device, and one for the switch/coil within the device.

## Switch/Coil Devices

For coils, we now have `ICoilableDevice` (item), `IApiCoilDevice` (API) and `ICoilDeviceAuthoring` (component), for switches it's `ISwitchableDevice`, `IApiSwitchDevice`  and `ISwitchDeviceAuthoring`. 

In terms of auto-population, two more fields were added to the switch/coil objects, in order to match the necessary elements for the default tables.

## Trough

For troughs, we now have a new GameItem with ID `101` that is read and written to the `.vpx` file. While the trough's runtime logic is in `TroughApi`, the switches and triggers are generated in the main `Trough` class based on its properties.

During runtime, the following happens:

1. The player initializes all playfield items (including switch/coil devices)
2. During `Init()`, the trough (which is both of the above) creates switch objects based on its configuration
3. The player sets up the switch and coil mapping. Now the devices have generated their switch objects, the player can link them to the gamelogic engine.
4. Game starts

The mentioned switch objects are instances of `DeviceSwitch`, which is a standalone implementation of `IApiSwitch`. It's basically a proxy where `OnSwitch()` can be called on and the event will be dispatched to the switch input(s) of the gamelogic engine.

Import: The converter automatically adds a trough when there is none. It also tries to bind it to `Drain` and `BallRelease`. The result is that importing the default table results with a working and fully set up trough.

Editor: In order to keep the inspector code DRY, the `SurfaceField` method has been refactored to be generic, so it supports the now necessary kicker- and trigger picker.

## Other Changes

- When an Item doesn't define its `ItemGroupName`, it's attached to the table root directly.
- Item APIs now have a `OnDestroy()` method they can override in case they need to unsubscribe from any events.
- Fixed a bug where rubber wasn't transformed. Fixes #234.

## TODO

- [x] Port actual trough logic from [core.vbs](https://github.com/vpinball/vpinball/blob/master/scripts/core.vbs#L251-L682)
- [ ] Check if we keep the jam switch and entry coil and discard if not
- [x] Proof-read documentation
- [x] Update switch- and coil documentation
- [x] Update switch- and coil unit tests
- [x] Refactor: Change components so we don't need to implement `IRenderable`
- [x] Fix: Ball Release doesn't always seem to work


ecurtz:
Did a sanity check on the documentation.

VisualPinball.Engine/VPT/Mappings/Mappings.cs
```diff
@@ -15,9 +15,12 @@
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.IO;
 using System.Linq;
 using System.Text.RegularExpressions;
+using NetVips;
```
jsm174, line 22: Are `NetVip` and `System.Diagnostics` really needed?

VisualPinball.Engine/VPT/Trough/Trough.cs
```diff
@@ -0,0 +1,69 @@
...
+		public override string ItemName { get; } = "Trough";
+		public override string ItemGroupName { get; } = null;
+
+		public const string JamSwitchId = "jam";
+		public const string EjectCoilId = "eject";
+		public const string EntryCoilId = "entry";
+
+		public IEnumerable<GamelogicEngineSwitch> AvailableSwitches => Enumerable.Repeat(0, Data.SwitchCount)
+			.Select((_, i) => new GamelogicEngineSwitch {Description = SwitchDescription(i), Id = $"{i + 1}"})
+			.Concat( new[]{ new GamelogicEngineSwitch{Description = "Jam Switch", Id = JamSwitchId} });
```
jsm174, line 37: `{` spacing doesn't seem to match. 

Const names don't really match how https://github.com/freezy/VisualPinball.Engine/blob/master/VisualPinball.Unity/VisualPinball.Unity/Game/Engine/DefaultGamelogicEngine.cs#L45 is done.

VisualPinball.Engine/VPT/Trough/TroughData.cs
```diff
@@ -0,0 +1,87 @@
...
+		public override string GetName() => Name;
+		public override void SetName(string name) { Name = name; }
+
+		[BiffString("NAME", IsWideString = true, Pos = 9)]
+		public string Name;
+
+		[BiffString("ENTK", Pos = 1)]
+		public string EntryKicker = string.Empty;
+
+		[BiffString("EXIT", Pos = 10)]
```
jsm174, line 44: Not really sure how `Pos` works, but is `6`, `7`, and `8` skipped for a reason?

jsm174 (approved):
Awesome PR. Just a few nitpicks, nothing crazy. Can not wait to try this with WPCEmu!

VisualPinball.Engine/VPT/Mappings/Mappings.cs
freezy: nope, cleaning.

VisualPinball.Engine/VPT/Trough/Trough.cs
freezy: I'll add the space :)

Don't understand your other remark about the const names. Can you elaborate?

VisualPinball.Engine/VPT/Trough/TroughData.cs
freezy: No reason.

The goal of `Pos` is to write the attributes in the same order than VP. It made binary comparisons easier when I was implementing it. They are just a sort key, so skipping isn't a problem, but I'll update them nevertheless.

VisualPinball.Engine/VPT/Trough/Trough.cs
jsm174: No big deal on this. Was just looking at `SwLeftFlipper`, `SwLeftFlipperEo`s, ie, those `const` names don't actually have the text `Id` in them.

freezy: Oh, okay. I guess it's a bit less obvious in the `Trough` itself than in `DefaultGamelogicEngine`, where pretty much all constants are IDs.

PR #245, opened 2020-11-17
Implement wire manager

### LINKED ISSUE #213: Add Wire Manager
While typically the gamelogic engine will hook up switches to triggers and lights where necessary, there are some cases where we would like to do it ourselves directly without having to implement a custom gamelogic engine.

The most obvious use case is to test out stuff, i.e. link coils or lights on the playfield to the keyboard. But I think there are also real-world use cases. For example I think there are games that have bumpers whose lamps are directly wired to their switches, i.e. the they are short-wired and don't go through the ROM.

So we should have a UI for connecting output elements (switches) to input elements (coils and lamps):

wire_manager

The description is again optional, then we have three blocks:
1. Source (switches)
2. Destination (coils and lamps)
3. Type and delay

Block 1 is pretty much the same as the switch manager. I'm not sure if the *constant* value makes sense here, but I guess testing lights could be done like that. Block 2 is like in the coil manager, only that it lets you select lamps as well (in the future, when we have a lamp manager, which is the same as the coil manager but for lamps).

Data-wise, it could look something like that:

```cs
public class MappingWireEntryData : BiffData
{
    public string Description = string.Empty;

    // source 
    public int Source = SwitchSource.Playfield;
    public string InputActionMap = string.Empty;
    public string InputAction = string.Empty;
    public string PlayfieldItemSource = string.Empty;
    public string DeviceSource = string.Empty;
    public int Constant;

    // destination
    public int Destination = WireDestination.Playfield; // use for both coil and lamp
    public string PlayfieldItemDestination = string.Empty;
    public string DeviceDestination = string.Empty;
    public string DeviceItem = string.Empty;

    // type
    public int Type = WireType.OnOff;
    public int Pulse = 10;
}
```

During gameplay, the player will then subscribe to the source events and trigger the destination elements. The relation between source and destination is n..n, i.e. the same element can trigger multiple destinations, and a destination can be triggered by multiple elements.

This issue is dependent on #211 and #212.

jsm174:
This PR closes #213.

We can keep this PR open as more development is still needed.

- Currently there is no lamp manager, so Destination + Destination Elements only support `ICoilable`
- `Type` and `Off` from https://github.com/freezy/VisualPinball.Engine/issues/213 have been modified to `Pulse Delay` based on `Source`
- Still need to code support into the Player

## TODO
- [X] Hook to player
- [x] Add documentation
- [x] Add test coverage


freezy:
Looks good! There's already an icon `plug` for the manager window. Will have a look at player integration if you want. `ICoilable` is fine, we'll handle lamps when we have them. 

Cheers!

Issue #244, opened 2020-11-15
Walls:  Add point has low success rate. 


Pandelii:
Adding points to wall lines has a high failure rate.  Out of 29 tries, it successfully created points 11 times.  It appears to succeed more often when the line of sight to the add location is unblocked by the wall geometry.  

Repro steps:  
Create wall, enter edit mode. 
Right click on line to add point.  

Expected result:  Point is added at current dot indicator on the line.  
Actual result: Point fails to be added in many cases.  

See attached video. 
[PointAddFailure.zip](https://github.com/freezy/VisualPinball.Engine/files/5542834/PointAddFailure.zip)

freezy:
Thanks!

@Vroonsh do you have time to take a look?

Issue #236, opened 2020-11-14
Bug: Wall editing,  point drag failure.


Pandelii:
When dragging points on a wall it is possible to produce a failure state by dragging one point directly over another.  

Repro steps: 
Create a wall
Enter point edit mode
Select the upper left point
Drag the red handle (x) to the right until it overlaps the upper right point. 

Expected result:  Nothing happens or a prompt appears to ask if you want to merge the points. 
Actual result:  The point disappears and the console log indicates a failure. 

```
NullReferenceException: Object reference not set to an instance of an object
VisualPinball.Unity.MeshExtensions.ApplyToUnityMesh (VisualPinball.Engine.VPT.Mesh vpMesh, UnityEngine.Mesh mesh) (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity/Extensions/MeshExtensions.cs:85)
VisualPinball.Unity.ItemMeshAuthoring`3[TItem,TData,TAuthoring].UpdateMesh () (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity/VPT/ItemMeshAuthoring.cs:138)
VisualPinball.Unity.ItemMeshAuthoring`3[TItem,TData,TAuthoring].RebuildMeshes () (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity/VPT/ItemMeshAuthoring.cs:91)
VisualPinball.Unity.ItemMainAuthoring`2[TItem,TData].RebuildMeshIfDirty () (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity/VPT/ItemMainAuthoring.cs:114)
VisualPinball.Unity.Editor.TransformInspector.RebuildMeshes () (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/TransformInspector.cs:125)
VisualPinball.Unity.Editor.TransformInspector.OnSceneGUI () (at E:/VPE/VisualPinball.Engine-master/VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/TransformInspector.cs:170)
[...]

PR #246, opened 2020-11-17
Drag points : various fixes

Vroonsh:
- fixed blocked handles on curve traveller right clic
- fixed control point insert index calculation
- fixed crash in surface mesh generator when crossing/overlapping control points

close https://github.com/freezy/VisualPinball.Engine/issues/244
close https://github.com/freezy/VisualPinball.Engine/issues/236


freezy (commented):
Tested and couldn't reproduce any non-working *Add Point*s anymore. 

Cheers!

freezy (approved):

PR #252, opened 2020-11-22
Added a few basic helper scripts.

Pandelii:
Added Table Manager - Provides access to references of existing tables in the scene.
Added Asset Paths - Provides paths to various editor resources.  Primarily used for prefab paths.
Added Editor Setup, Blueprint project and Hierarchy GameObject Menu (stub in).  (requires asset library package to function, but not to compile)
Added VPE Render Pipeline definition script to allow for ifdef code blocks based on pipeline.


freezy:
All merged into the repositories now. Thanks @Pandelii and sorry it took so long!

PR #258, opened 2020-12-02
Unity: Refactor all native plugin meta files

jsm174:
This PR is to address errors seen when trying to compile Windows native builds. 

We submitted a bug report to Unity, and posted it here:

https://forum.unity.com/threads/build-standalone-cross-platform-linux-error-multiple-precompiled-assemblies-with-the-same-name.961458/

Even though the Unity inspector reflects the desired settings, it is actually using other entries from the meta files:

Screen Shot 2020-12-02 at 5 32 01 PM

In this example, exclude `Win` and `Win64` were set to `0`, thus they were still being including, causing the `Multiple precompiled assemblies` errors.

All native plugin meta files were gone through by hand and all the extraneous settings were removed.

### DIFF
--- a/VisualPinball.Unity/Plugins/linux-x64/NetMiniZ.dll.meta
+++ b/VisualPinball.Unity/Plugins/linux-x64/NetMiniZ.dll.meta
@@ -16,25 +16,14 @@ PluginImporter:
     second:
       enabled: 0
       settings:
-        Exclude Android: 1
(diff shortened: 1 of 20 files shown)


freezy:
Nice! This means builds now work out of the box?

jsm174:
> Nice! This means builds now work out of the box?

yup!

PR #260, opened 2020-12-06
Replaced manual material changes with material default templates

Roland09:
Created material templates of which the properties are copied to the new created material. This is more future proof and supports automatic upgrade via Unity's own mechanism in case the shader changes.


freezy:
Thanks! 👍

PR #261, opened 2020-12-07
Move TableManager from #252 to EditorTable.

freezy:
Start small I guess. ;)

I didn't like the name `TableManager`, because all our managers are editor windows. So I renamed it to `EditorTable`. If you have a better idea, let me know! 

There were also a few style changes, e.g. I prefer putting even oneliners into braces because they facilitate diffs and provide better readability. I also try to use common C# naming standards except for namespaces and serializable fields where we use Unity's.

I've also removed `#if UNITY_EDITOR`, because it's part of the editor project which won't be compiled into the binary when building.

No (end-user) documentation needed for this one.

This is the first PR of a series in order to get #252 upstream.


Pandelii (approved):
Looks good to me.

PR #267, opened 2020-12-29
Remove ILRepack to prep for VisualPinball.Engine in NuGet 

jsm174:
In order to publish `VisualPinball.Engine` as a NuGet, we need to make sure all its dependencies can also be found in NuGet. 

`NetVips` has dependencies on `System.Memory`, `System.Buffers`, and `System.Runtime.CompilerServices.Unsafe`. These dependencies were colliding with dependencies required by `Unity.Collections`. This [post](https://forum.unity.com/threads/please-update-com-unity-collections-reference-of-system-runtime-compilerservices-unsafe.732056/#post-5972735) has more details. 

Because of this, we had to use `ILRepack` to make a "custom" version of `NetVips` that had all the dependencies built in. 

A NuGet version of `VisualPinball.Engine` would not have access to this "custom" version of `NetVips`.

Now that we are using `Unity.Collections 0.14.0-preview.16`, we do not need to use `ILRepack`. We do though, need to include the `System.*.dll`'s in all the plugin platform folders -- they need to be in the same folder as `NetVips.dll`.

This PR also makes sure all the Unity meta files are properly included/excluded per platform.


freezy (approved):
Looking good. Thanks for this!

PR #268, opened 2020-12-31
Update NetMiniZ to prep for VisualPinball.Engine in NuGet

jsm174:
In order to publish `VisualPinball.Engine` as a NuGet, we need to make sure all its dependencies can also be found in NuGet.

`VisualPinball.Engine` has a dependency on `NetMiniZ`.  `NetMiniZ` was using a packaging strategy that was different than `NetVips`. This was adding extra unnecessary files to the `VisualPinball.Engine` nuget.

This PR updates `NetMiniZ` to 1.2.2 which uses the same packaging strategy as `NetVips`.


freezy (approved):
Awesome, thanks a lot!

PR #270, opened 2020-12-31
ci: add release workflow and support for publishing to nuget

jsm174:
This PR adds support for publishing to NuGet. 

We reworked the `Publish` workflow into `Release`, and then made a new `Publish` workflow the publishes to the Visual Pinball Registry and NuGet.

The github workflows will work as follows:

```
documentation
build -> release -> publish (registry / nuget) -> dependents
```

If desired, we can adjust to:

```
documentation
build -> release -> publish (registry / nuget)
                 -> dependents
```

Unfortunately, we can not actually test the NuGet publish process until this is merged into master.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,6 @@ jobs:
       - uses: actions/setup-dotnet@v1
         with:
           dotnet-version: '3.1.x'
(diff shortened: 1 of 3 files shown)


freezy (approved):
Let's try this!

PR #276, opened 2021-01-03
Populate editor layouts

freezy:
Part of #252, this ports `PopulateEditorLayoutMenu.cs` and `LayoutUtility.cs`. The additional resolution options have been dropped since Unity 2020.2 already comes with a full HD and 4K preset. Might be worth re-adding for portrait support later though.

The actual layout files have moved into this repo, since we don't have the asset library dependency in the main repo yet (and I'm reluctant to add it until we really need it). So assets have been removed from then asset library. I've also removed the asset paths to the layouts and hard-coded it in the class. I think that's the better approach since it's really specific to the class and not used elsewhere.

Finally, if you've tested #252, and you already have the layouts, you'll have to manually delete them (at `%AppData%/Unity/Editor-5.x/Preferences/Layouts/default`) or have duplicates because there is an additional layout.

## TODO

- ~[X] Complete documentation.~ Will be mentioned in the setup guide once the HDRP project is set up correctly.


Pandelii:
As we discussed:  

Two things on the layout tool that we should consider altering at some point:  
1) Initialization behavior feels like it may happen too often.  Should be more of a one time process.  
2) We should consider adding a log event if the API changes or fails.

freezy:
Good point. It's now a menu and I could also get rid of the reflection code, since we don't use it anymore.

freezy:
How the menu looks now:

image

PR #283, opened 2021-01-19
Gamelogic Engine updates

freezy:
- Merged `IGamelogicEngineWithCoils`, `IGamelogicEngineWithSwitches` and `IGamelogicEngineWithLamps` into `IGamelogicEngine`
- Removed lifecycle methods of `IGamelogicEngine`, because they are usually MonoBehaviours

That's it!


jsm174 (approved):
This really makes the gle a lot cleaner!

freezy:
Yeah, I can't remember who, but it was already mentioned why the split, and there isn't any reason actually.

freezy:
Thanks for the review!

PR #284, opened 2021-01-21
Update NLog to 4.7.7

jsm174:


freezy (approved):
Thanks! Also thanks @snakefoot for reacting so quickly!

PR #285, opened 2021-02-01
More Gamelogic Engine Refactoring

freezy:
Working on PinMAME and MPF there were a few things we had to change, namely:

- We have IDs documented in the game manuals, and IDs that PinMAME is using. They aren't always the same. So we've added an integer ID field to the mappings which is the actual value sent to PinMAME (or any other GLE).
- We've also added a **Normally Closed** boolean field to the mapping which inverts the value sent to the gamelogic engine if set. This means that most of our switch methods don't deal with "open" and "closed" anymore, but with "enabled" and "disabled". The actual "closed" value sent to the gamelogic engine is computed by the player before sending.
- Switch devices can additionally hint whether their switches are all NC, NO, or configurable. This is to properly render status in the editor window (e.g. a opto-switch trough would have its switches NC).
- Coils can also be lamps (typically flashers). Setting a coil destination to *lamp* will add a new mapping to the lamp manager where it can be configured accordingly.
- Added the DMD shader to the global assets.

## TODO

- [x] Figure out how to deal with flippers. Before this PR, we fully handled dual-wound flippers, which was the default behavior. Now, PinMAME seems to handle all flippers as single-wound. So either update the default gamelogic engine's flipper handling, or dynamically set the flipper type.
- [X] ~Handle the plunger correctly. I think it'll need two inputs: The autoplunger that kicks the ball directly, and the "start pulling" input of the manual plunge. So it should probably be a switch device. At some point this will be entirely analogue though, so I'll probably create a new issue for this~ #286.
- [x] Test coverage


jsm174 (commented):
This is great.

One comment, did you mean to include the DMD material/shader/meta files in this PR?

freezy:
> did you mean to include the DMD material/shader/meta files in this PR?

Yes, what do you think is missing?

jsm174:
Nothing. Sorry, it just seemed out of place. 

Was thinking there would be a DMD PR someday and it would be a part of that.

freezy:
Yeah I just included it for debug purposes. Will do a proper component in another PR at some point.

PR #293, opened 2021-03-12
MPF / Add support for dynamic wires

freezy:
Dynamic wires, also known as "hardware rules", are wires the gamelogic engine can create or remove during gameplay. It does that by linking a switch to a coil, where both the switch and the coil must be mapped through the switch manager and the coil manager respectively.

The first gamelogic engine to make use of this is MPF. The goal is to eliminate latency between the switch and the coil, which would otherwise result by processing switch events through software, which runs at a much lower frequency.


freezy:
@ecurtz @Pandelii or @jsm174, any way one of you could proof read the doc? Basically those four files:

-  https://github.com/freezy/VisualPinball.Engine/blob/feature/mpf/VisualPinball.Unity/Documentation%7E/plugins/index.md
- https://github.com/freezy/VisualPinball.Engine/blob/feature/mpf/VisualPinball.Unity/Documentation%7E/plugins/mpf/index.md
- https://github.com/freezy/VisualPinball.Engine/blob/feature/mpf/VisualPinball.Unity/Documentation%7E/plugins/mpf/setup.md
- https://github.com/freezy/VisualPinball.Engine/blob/feature/mpf/VisualPinball.Unity/Documentation%7E/plugins/mpf/usage.md

freezy:
It's the MPF doc, feel free to push directly to this branch.

PR #296, opened 2021-03-31
added "open the *Other Settings* tab,"

ScottKirvan:
The "Scripting Define Symbols" was hidden within that tab


freezy (approved):
Awesome, thanks a lot!

PR #301, opened 2021-05-22
Added Goldorak Patcher

Roland09:
Added a new patcher for Goldorak. Repareting the Flippers so that they move


freezy (approved):
Oops sorry missed that one completely.

Thanks!

PR #302, opened 2021-05-30
Refactor Scene Setup

freezy:
This is still WIP, but I'll write some documentation here so people can validate before I go into too deeply (although the ground work is already done).

Before this PR, importing a VPX file would write the entire table data into the scene. This approach had several issues:

- Inefficient: When we extract heavy data like textures and meshes, the original data was still kept, and very badly: While Unity stores meshes as serialized binary data, our data meshes were serialized as objects, resulting in gigabytes of scene data for medium-sized tables.
- Synchronization problems: Because game play, export etc were still using the originally imported `Table` object, hacks were necessary to update this object when stuff in the scene was added or removed. It actually didn't really work at all.
- Not editable: Assets like textures and sounds couldn't be edited or replaced, because they were still stored as serialized objects.

## Table Container

To solve the synchronization issue, we've split the `Table` class, which was the root object of all the data, into two classes:

- `Table` now only holds the "table data", which are the global parameters of the table, but no game items, textures, or anything else.
- `TableContainer` is an abstract class that provides access to the rest of the data, which are textures, materials, sounds, game items, collections, global table props, custom info tags, and mappings.

Now, when the table file is read, we get a instance of `FileTableContainer`, a subclass of `TableContainer`. That's how we get access to all the data in the file. That data is then converted to a scene, where we have a `SceneTableContainer` (as a field of `TableAuthoring`). This container also provides access to all the data, but it pulls it out of the scene. This means we don't need to sync anymore, since the old `FileTableContainer` doesn't exist anymore once import has finished.

`SceneTableContainer` doesn't contain any state, it just knows where to retrieve the state from. For example, the binary data of a texture would be retrieved from an image in the asset folder (see below), or the metadata of a flipper would be retrieved from the flipper's `FlipperAuthoring` component, which still sits in the scene.

We currently refresh the references of `SceneTableContainer` on `EditorApplication.hierarchyChanged`, but this might not be necessary, maybe we can lazy-refresh only when we actually need fresh data.

## Assets

When importing, assets are now written to separate files in the asset folder whenever possible, to keep as little as possible in the scene.

- Textures are written as image files
- Sounds are written as .wav files
- Materials are saved as .mat files
- Physical materials are saved as .asset files
- Primitive meshes are saved as .mesh files, generated meshes are not saved
- All game items are saved as prefabs

This means that we can now edit textures directly on the disk and Unity will automatically re-import them.

Note that we still store all the original data in the authoring component of the game object. However, we "free" heavy, fields once extracted. For example, after the mesh data of a primitive is saved as an asset, the mesh of the data object is nulled so it doesn't add redundant data to the scene.

### Materials

Materials from Visual Pinball are saved for later export, but are ignored otherwise. That means updating them doesn't change anything in the converted Unity material, and changing the converted Unity material doesn't update the original material. If a material is used as physics material, or if it has the physics properties set, the material is saved as asset as a `PhysicsMaterial` object and linked to the collision authoring component. During gameplay, the asset is used instead of the table data.

## Side Car

The "side car" object was initially used to off-load heavy data into a separate `ScriptableObject`. I don't see any use of this anymore. However, we still would like to re-export data we currently don't use in VPE. Thus, this object was renamed to `LegacyContainer` and contains data that is in the VPX file, but ignored by VPE.

## TODO
- [x] Implement `SceneTableContainer.GetMaterial()`
- [x] Implement `SceneTableContainer.GetTexture()`
- [x] Handle asset folder name:
  - [x] When importing, use file name instead of table object name
  - [ ] ~When creating a new table, ask for a name~ (Table1 is fine, it can be renamed)
- [x] Verify legacy data is saved correctly
- [x] Handle the storage indices, probably best to re-index just before exporting, since we don't need them anywhere else.
- [x] Handle imported meta data of textures, sounds and materials when exporting
- [x] Don't re-import assets that already are in the asset folder
- [x] Test other sound formats than .wav
- [ ] ~Convert the "resources" dependency to prefabs~ (new issue)
- [ ] ~When exporting, add an option to automatically include all meshes in the scene as primitives (non-collidable toys), even when no `PrimitiveAuthoring` component is set.~ (new issue - maybe)
- [x] Fix test coverage


freezy:
To fix coverage, we're now running unit tests through Unity. This means we can also unit-test importing the table into the scene and exporting it back to `.vpx`, which is pretty cool!

freezy:
Few more things before merging:

- [ ] Make `WRITE_VP106` and `WRITE_VP107` settings instead of compiler flags
- [ ] Move non-VPX game items to another stream name so they stack backwards compatible
- [ ] Cover webp and ogg in unit tests
- [ ] Test more real-world tables

PR #304, opened 2021-06-11
Runtime camera controller update and new clipping component. 

Pandelii:
Added runtime camera controls to the camera controller.  Removed old camera clip plane adjustment setup.  Added new stand-alone camera clip plane component to allow alternate camera controllers.


VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+using System; 
+using Unity.Mathematics;
+using UnityEngine;
+
+
+namespace VisualPinball.Unity
+{
+	[ExecuteAlways]
```
freezy, line 24: This means the life cycle is applied in the editor as well, right? What's the editor function of this MonoBehavior? I see that below you're fetching the table bounds differently whether you're in play mode or not, but the camera itself is only active in play mode, no?

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+		{
+			SetClipPlanes(0.001f, 9f); 
+		}
+
+		private void Update()
+		{
+			if(!GetCameraComponent())
+			{
+				return;
+			}
```
freezy, line 45: So this searches for the camera on every frame?

Wouldn't it be better to search for it only once in `Awake()`?

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+			{
+				Camera = UnityEngine.Camera.current;  //Get the current active camera if not on the camera component.  
+			}
+
+			if(Camera == null)
+			{
+				return false;
+			}
+
+			return true;
```
freezy, line 122: Faster to read :)

```suggestion
			return Camera != null;
```

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+		
+		/// <summary>
+		/// Sets the camera clipping planes based on manually derived values. 
+		/// </summary>
+		/// <param name="near">The near clip plane value</param>
+		/// <param name="far">The far clip plane value</param>
+		/// <returns>False when no camera could be found.</returns>
+		public bool SetClipPlanes(float near, float far)
+		{
+			if(Camera == null) return false;
```
freezy, line 133: I prefer adding the curly braces because it makes diffing easier, it's more explicit to read, and makes it less error-prone when you need to add more lines to the block (not likely in this example though).

```suggestion
			if (Camera == null) {
				return false;
			}
```

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+			SetClipPlanes(nearPlane, farPlane); 
+
+			return true; 
+		}
+
+		/// <summary>
+		/// Gets the currently active or attached camera component. 
+		/// </summary>
+		/// <returns>False when no camera can be found.</returns>
+		private bool GetCameraComponent()
```
freezy, line 108: Maybe I have done too much Java in the past, but "Get" methods usually indicate a quick operation (e.g. returning an already set member). Here you're doing `GetComponent()` which is more expensive, so I'd rather call it `RetrieveCameraComponent()`.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+			return true; 
+		}
+		
+		/// <summary>
+		/// Sets the camera clipping planes based on manually derived values. 
+		/// </summary>
+		/// <param name="near">The near clip plane value</param>
+		/// <param name="far">The far clip plane value</param>
+		/// <returns>False when no camera could be found.</returns>
+		public bool SetClipPlanes(float near, float far)
```
freezy, line 131: The return value isn't used, so you can make it `void`!

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
```diff
@@ -0,0 +1,142 @@
...
+			{
+				UpdateClipPlanes();
+				previousCameraPosition = Camera.transform.position; 
+			}
+		}
+
+		/// <summary>
+		/// Updates the camera clipping planes based on the table bounds.
+		/// </summary>
+		public bool UpdateClipPlanes()
```
freezy, line 57: Return value isn't used.

freezy (commented):
Thanks, this is great. I've added a few comments, but I couldn't test it yet, so I'll have a look at `CameraController` later tonight. Cheers!

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: The game view panel rendering is used in both modes so if the camera is moved it needs to update the clipping planes for that view.  The camera is always active.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: Yeah, that would be better and should be fine.  A null check before updates would be sufficient to avoid nres

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: Good call.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: It isnt used at the moment, return value provided in case other camera controllers that people use or make need to know if it was successfully set since it can fail. Not essential by any means though.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: Sounds good.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
freezy: Got it, thanks!

No serious performance percussions I assume?

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: Ended up doing it slightly differently.  It will search for a camera every frame until it has one because it doesn't appear to be accessible with Awake in all cases.

VisualPinball.Unity/VisualPinball.Unity/Game/CameraClipPlane.cs
Pandelii: Nah, it's a pretty cheap check.  Just polling for a location change and if it notices one, it updates otherwise it skips.

PR #305, opened 2021-06-11
Add nFozzy Flipper Corrections

syllebra:
This is a first raw implementation of nFozzy's flipper corrections: polarities and velocities.
It is improved by using Unity's "Animation curves". Some preset cruves are still missing.

image


djrobx:
Very nice - happy to see this included!

freezy:
Me too 😄

BTW, welcome back, Rob!

VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperCorrection.cs
```diff
@@ -0,0 +1,151 @@
...
+			if (L < 0)
+				return defaultValue;
+
+			if (xInput > curve[curve.Length - 1].x) // catch line overrun
+				L = curve.Length - 1;
+
+			float y = PSlope(xInput, curve[L - 1].x, curve[L - 1].y, curve[L].x, curve[L - 1].y);
+
+			return y;
+		}
```
freezy, line 131: This one doesn't seem to be used? Ditch in favor of the next method?

syllebra: This one was the direct port of nFozzy's code. Th eother one is an optimization only valid if the segments of the curve are evenly distributed on x axis. So it will depend on what we plan for the curves to come. It is indeed not used right now as the Blobs are filled evenly splitting the aniamtioncurves.

freezy: Ok, keeping it for now then.

freezy:
I've removed `FlipperCorrectionAuthoring` and put all the data into `FlipperCorrectionAsset` (formerly `AnimationCurveAsset`). IMO this makes it easier for authors to handle profiles. A profile is basically the asset.

So it can now be enabled under in the collider component:
image

And edited in the inspector:



Selection is easier as well:


I'll add the documentation and after that it should be good to merge.

freezy (approved):

freezy:
Thanks to all involved, merging.

PR #306, opened 2021-06-14
misc: fix trough optos to be normally open

jsm174:
- When a ball is sitting between an opto, it breaks the light beam, ie "opening" the switch (0 - false)
- When a ball is not between the light beam, the light beam is passing through, ie "closing" the switch (1 - true)

That said, PinMAME treats open Opto switches as Active (true). This is the reverse of the above.

Screen Shot 2021-06-14 at 4 36 41 PM

A better explanation can be found [here](https://pinside.com/pinball/forum/topic/lets-discuss-williams-opto-theory#post-2424066):

> Part 2 - Why optos are ACTIVE when they're blocked..
> 
> The IR reciever (Phototransistor) allows the row circuit to be closed (i.e. switch closed, current flowing from collector to emitter) when it detects IR light. When there is no IR light detected, the phototransistor is open, stopping current from flowing.
> 
> Since the ball blocking the IR beam is activating the switch, this is why in switch test that -
> 
> An ACTIVE switch is one that is not in its normal state. For regular switches, this is when the switch is closed, current flowing. For optos, this is when the switch is open (the IR beam is broken, switch is open).
> 
> This is also why, when running switch tests, and looking for help on here, you need to pay attention to whether the switch is Active - has an "(A)" after the name/status) or simply 'On'. 'On' is a misnomer, since optos are supposed to be on, other switches aren't.

This PR reverses the defaults for Opto switches.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
@@ -197,7 +197,7 @@ namespace VisualPinball.Unity
 				_stackSwitches = new DeviceSwitch[Data.SwitchCount];
 				foreach (var sw in Item.AvailableSwitches) {
 					if (sw.InternalId > 0) {
-						_stackSwitches[sw.InternalId - 1] = CreateSwitch(sw.Id, false, Data.Type == TroughType.ModernOpto ? SwitchDefault.NormallyClosed : SwitchDefault.NormallyOpen);
+						_stackSwitches[sw.InternalId - 1] = CreateSwitch(sw.Id, false, SwitchDefault.NormallyOpen);
 						_switchLookup[sw.Id] = _stackSwitches[sw.InternalId - 1];
 					}
 				}
@@ -209,7 +209,7 @@ namespace VisualPinball.Unity
 			}
 
 			if (Data.JamSwitch) {
-				JamSwitch = CreateSwitch(Trough.JamSwitchId, false, Data.Type == TroughType.ModernOpto ? SwitchDefault.NormallyClosed : SwitchDefault.NormallyOpen);
+				JamSwitch = CreateSwitch(Trough.JamSwitchId, false, SwitchDefault.NormallyOpen);
 				_switchLookup[Trough.JamSwitchId] = JamSwitch;
 			}
 


freezy:
Thanks. I'm a bit hesitant though, because this will change the behavior for other gamelogic engines as well, and the trough should be machine agnostic.

But I'm not sure which is the least confusing way to handle this. Here's the different solutions I can think of:
1. Assume most authors will run PinMAME and just make it work with PinMAME (this PR)
2. Assume authors know about the normalized PinMAME values, and will always choose a mechanical trough (would be my preferred solution)
3. Add a "does normalize optos" flag to the gamelogic engine, and make the player invert accordingly (this is a bit harder because we don't have a reference to the gamelogic engine in the trough - it just emits events, so we'd need to include in the event whether it's an opto or not).

Thoughts?

jsm174:
Yeh, I started thinking the same thing as I typed out the PR. 

I personally like #3, however:

I'm guessing there have to be devices other than troughs that have opto's. 

What if in the switch manager, a switch had a "type" column, ie, mechanical, opto? Then when populating hardware, the switches could be set accordingly. 

If we didn't want the GLE to include a normalize opto's setting, you could even break the type columns down further to Opto (Active open), Opto (Active closed)?

freezy:
We already have this, it's the NC column (here with a modern opto switch):

image

The problem is that if we let the user edit this for optos, it would defeat the purpose of the opto trough in the first place (it basically makes it a mechanical opto again).

I think 3 is alright as well. When you include the "NC" prop in in the event, we don't care from where it comes. Could be other items than troughs. Then the PinMAME GLE would just invert NC signals. However, if there were other NC switches on the table that are *not* normalized, then the author would need to invert them in the switch manager to PinMAME reads the correctly (unless, of course, we add another flag to the event, `IsTrough`, so we could filter).

This will be documented of course. The trick is to find the least confusing variant, and preferably keeping the confusing parts for the least likely setups.

jsm174:
Was just thinking too, IJ uses opto's for the flippers, and we never saw issues there (because we never did anything different with that logic) 

Screen Shot 2021-06-14 at 6 07 06 PM

freezy:
Yeah that would speak for solution 2. Don't even need to know they're opto if PinMAME interprets them as mechanical anyway.

jsm174:
Okay, so I guess we can close this PR then? 

I was going to start trying to work on 2 by modifying PinMAME's GLE,

Screen Shot 2021-06-15 at 7 51 49 AM

but then we would need to get to the `SwitchPlayer` / `SwitchStatusesClosed` so that initial switches could be set properly. And I think that would really ugly up the code.

I switched MM to use mechanical, and it's working.

freezy:
Yeah, given simplicity comes into play as well (it's already not trivial now, took me a while to understand how NC/NO works), so I'd prefer to put a note about optos in the PinMAME doc.

jsm174:
Thanks!  Closing this PR as not needed.

PR #311, opened 2021-06-17
patcher: add support for cftbl

jsm174:
This adds support for Creature From The Black Lagoon.

It reparents the flippers, removes the flipper shadows,  and fixes the angle and speed for entry and exit kickers.


freezy (approved):
Thanks!

PR #312, opened 2021-06-18
Camera Controller Bug Fix and Plunger Fix

Pandelii:
Gated the active setting grab until the application is playing.

Added DOTS_INSTANCING_ON keyword and corrected missing guid for lerp vertex subgraph on plunger shader.


freezy (approved):
Thanks!

PR #313, opened 2021-06-30
Add new ID when it does NOT exist.

freezy:
Seems like an obvious bug, not sure why it was there in all three managers.

Thanks @scottacus for reporting.


jsm174 (approved):
Looks good!

PR #317, opened 2021-08-02
Properly use prefabs

freezy:
Currently we save every game item as prefab to the asset folder. Since the data sits in a different object, updates via [SerializedProperty](https://docs.unity3d.com/ScriptReference/SerializedProperty.html) aren't possible, and thus we can't easily override the prefabs.

This PR copies the data into the components, where we can more easily override them. Having them in the component additionally allows us to properly save references to other game items (e.g. surfaces) through a references instead of a string, and also allows us to only pick VPE relevant properties.

Additionally, every game item is now instantiated from an existing prefab that we ship ourselves. That means one prefab per game item, or more precisely, one prefab per game item variation. Sub types would be different plate meshes for gates, or primitive flipper meshes with proper collision set up. Game-specific data would result in overrides.

The goal is to remove the `Data` objects from the authoring classes. It's done for the API classes already.

## TODO:

- [x] Fix primitive meshes (specially mesh inspector)
- [x] Fix playfield mesh inspector
- [x] Move Trough tests to Unity project
- [x] ~Ditch `VisualPinball.Resources.dll`~ Exported `.mesh` files have different vertices now, will debug another time...
- [x] Make lights Wire Manager compatible
- [x] Deal with table's `Data`
- [x] Find a way to deal with mesh variations (replace mesh? replace prefab?)
- [x] Make tests pass
- [x] Verify that material- and image manager still makes sense, otherwise drop it
- [x] Refactor switch manager with references
- [x] Refactor coil manager with references
- [x] Refactor lamp manager with references
- [x] Refactor wire manager with references
- [x] Move mapping tests to unity project
- [x] Verify that patching on import still works (pre and post)
- [ ] Update doc / screenshots / class schemas where necessary
- [x] Remove other unnecessary props (`ItemMainAuthoring.IsCollidable`, etc)
- [x] Get rid of `Table` if possible
- [x] Move ramp surface height implementation into Component
- [x] Make updating `ISurface` update items linked to that surface as well (so everything moves with the surface)
- [x] Remove unneeded `[MaterialReference]` and `[TextureReference]` from datas (made obsolete)
- [x] Add missing prefabs to built-in pipeline
- [x] Verify URP prefabs are all available
- [x] Fix "Force Update Mesh" button (should work when reset, and when mesh comp was added manually)
- [x] Check it `TableAuthoring.TableHeight` is persited after play mode
- [x] Dragpoints: Sometimes, don't show them. like for triggers with circle collider, and there was another use case.
- [x] Handle custom playfield mesh
- [ ] Refactor `GetRenderObject` to make it more clear which are used to create meshes, and which to only update meshes (basically only return what's needed, as opposed to currently always returning the render object group).
- [x] Check if we can remove `IItemAnimationAuthoring`
- [x] Check if we can remove `IItemColliderAuthoring` and interfaces in `VisualPinball.Unity` (VPT) in general
- [x] Fix layer management
- [x] Fix rubber collision mesh (height)
- [x] Fix trigger collision mesh (height)

### For each game item:

- [x] Change player startup to use Start() instead of Awake() to make components toggable
- [x] Abstract `Data` to `ISomethingData` etc to avoid re-instantiating data and use the component directly instead
- [x] Ditch `Data` and `Item` from authoring components
- [X] Remove `Data` and `Item` from API classes
- [x] Rename `Authoring` to `Component`
- [x] Check if data is split up correctly.
  - [x] Decouple dependencies: An item could (by script) animate even if there is no collider
- [x] Verify that creation via toolbox still works
- [x] Verify editor tools (move, rotate, scale) still work
- [x] Verify there is an easy way to change mesh when there are multiple prefabs
- [x] Verify visibilities, import and export
- [x] Verify colliders. Both in editor debug and runtime
- [ ] Verify materials. Import ~and export~
- [x] Verify undo works on all actions
- [x] Verify drag point edit works
- [X] Verify Extension class is in editor project and namespace
- [x] Verify Tooltips are set for all props
- [x] Verify Surface position works
- [x] Check `GetComponent<>()` vs `GetComponentInChildren<>()`

### Optionally

- [x] Ditch `table.GetScaleZ()` everywhere


Roland09:
For me the main goal is to move the data out of the scene into assets of the project folder in order to make them for one reusable and for the other to keep the scene file small. Previously the scene file had 3 GB for certain tables. Now it's way better, also the play mode starts a lot faster than previously. Saving the scene doesn't take minutes anymore.

One may as well stick with gameobjects in order to keep it all flexible, given that the gameobject data like texture, mesh etc are assets of the project and not embedded in the scene.

freezy:
@Roland09 yeah, getting stuff out of the scene was needed indeed. I'm currently working on 1) and it looks promising. Converted bumpers and am currently on walls. I think we'll ditch the entire data object as well as the items from the scene and work with components only. It simplifies a lot of things, and should be easier for Unity people to dive in, since we're making full use of the existing editor features.

If it's not flexible enough we can always copy the package prefabs to the project folder during import so the user can edit them directly instead of having to override everything. That's relatively easily done though, the main work is ditching the data objects and reworking mesh generation and the transform restrictions we have.

A lot of things are broken right now though. Once I'm through with all the game items, I'll bugfix and review the code again, and then I hope to be done with refactoring for a while :)

PR #337, opened 2021-10-30
Add basic support for rotation mech

freezy:
Mechs are a complex topic and this PR starts by splitting the former cannon component in two:

- A `StepRotatorMechComponent`, which simulates the mech
- A `RotatorComponent` which applies rotation on the playfield, inclusively balls.

The rotator component ([documentation]( https://github.com/freezy/VisualPinball.Engine/blob/fix/teleporter/VisualPinball.Unity/Documentation~/creators-guide/manual/mechanisms/rotators.md)) works by introducing a new `IRotatableComponent` interface, which allows setting a z-rotation and retrieving the position on the playfield. One of those is the defined as the "target", and other objects can be added which will rotate around the target. Rotation is applied to the mesh, but also to other relevant attributes like the eject angle for kickers (however, colliders still cannot be moved). Balls in the kickers are correctly rotated as well.

It somewhat works now with a cannon, although there are many more changes needed in order to support more generic mechs.

https://user-images.githubusercontent.com/70426/139556198-e9eed966-0c88-4cf9-874f-f33de4137144.mp4

## Other Changes

- Fixed a bug in teleporter
- Bumped Unity to 2021.2
- Removed obsolete angle/speed from kicker (it's handled by the coils array)
- Don't spam console when using ball roller with no ball
- Fixed a bug when an object sits on a ramp

## TODO
- [x] Icons for rotation and step rotation components
- [x] Documentation for `RotatorComponent`


VisualPinball.Unity/VisualPinball.Unity/VPT/Kicker/KickerCollider.cs
```diff
@@ -100,9 +100,11 @@ internal static class KickerCollider
 							// Only mess with variables if ball was not kicked during event
 							ball.Velocity = float3.zero;
 							ball.AngularMomentum = float3.zero;
-							var posZ = staticData.FallThrough
-								? staticData.ZLow - ball.Radius - 5.0f
-								: staticData.ZLow + ball.Radius;
+							var posZ = !staticData.FallIn
+								? staticData.ZLow + ball.Radius * 2
+								: staticData.FallThrough
+									? staticData.ZLow - ball.Radius - 5.0f
```
jsm174, line 106: Is it worth making 5.0f a constant?

jsm174 (approved):
Awesome stuff! Looks good.

VisualPinball.Unity/VisualPinball.Unity/VPT/Kicker/KickerCollider.cs
freezy: It's like that in the VPX code, so I'd prefer to stays as close as possible.

PR #340, opened 2021-11-11
Add support for matching multiple devices in Switch/Coil managers

jsm174:
This PR adds support for matching multiple devices for "Populate All" in the Switch, Coil, and Lamp managers.

While working on Rock, we noticed that the upper and lower flippers are controlled together, using the same coil reported by PinMAME. This could be configured manually in the Coil Manager, but since this is a table I test with often, I wanted it done automatically:

Screen Shot 2021-11-11 at 4 47 33 PM

Screen Shot 2021-11-16 at 10 01 40 PM

VisualPinball.Engine.PinMAME will require an update as well:

```
new GamelogicEngineCoil(CoilFlipperLowerRight, 46) { Description = "Right Flippers", DeviceHint = "^(Upper|Lower)RightFlipper$", NumMatches = 2 },
new GamelogicEngineCoil(CoilFlipperLowerLeft, 48) { Description = "Left Flippers", DeviceHint = "^(Upper|Lower)LeftFlipper$",  NumMatches = 2 },
```

https://github.com/VisualPinball/VisualPinball.Engine.PinMAME/pull/16 needs to be merged after this PR. 

`NumMatches` limits the amount of regex matches.

This PR also updates dependencies such as `NetMiniZ`, `NetVips`, and `NLog`. It updates all the projects to `netstandard2.1` and adds support for `osx-arm64`.

Remaining:

- [x] Handle lamps the same way
- [x] Fix classic single ball trough
- [x] Add emission node to segments
- [x] Add slingshots for Rock
- [x] Add GI and light groups for Rock
- [x] Fix lamp intensities and fade times
- [x] Remove extra display lights for Rock
- [x] Add support for `osx-arm64`
- [x] Add Lamp manager to layouts 
- [x] Add Left / Right Advance for System 80B
- [ ] Bump to netvips native `8.12.0` (currently using `8.12.0-rc1`)


xantari:
@jsm174 Doctor Who would be another one that is probably this way. The left lower and left upper flipper fires at the same time. Though not sure how PinMAME reports that....

jsm174:
> @jsm174 Doctor Who would be another one that is probably this way. The left lower and left upper flipper fires at the same time. Though not sure how PinMAME reports that....

Cool. Do you want to work on a Doctor Who config file like this:

https://github.com/VisualPinball/VisualPinball.Engine.PinMAME/blob/master/VisualPinball.Engine.PinMAME/Games/Rock.cs

We could get it in a PR when I get the changes for Rock in.

xantari:
> > @jsm174 Doctor Who would be another one that is probably this way. The left lower and left upper flipper fires at the same time. Though not sure how PinMAME reports that....
> 
> Cool. @xantari: Would you want to work on a Doctor Who config file like this:
> 
> https://github.com/VisualPinball/VisualPinball.Engine.PinMAME/blob/master/VisualPinball.Engine.PinMAME/Games/Rock.cs
> 
> We could get it in a PR when I get the changes for Rock in.

Yeah that sounds fun might be able to look at it in the next week or two

VisualPinball.Unity/VisualPinball.Unity/Mappings/MappingConfig.cs
```diff
@@ -139,23 +168,23 @@ private static ISwitchDeviceComponent GuessSwitchDevice(ISwitchDeviceComponent[]
 
 				var swMatch = switchDevices.FirstOrDefault(sw => sw.name == matchKey);
 				if (swMatch != null) {
-					return swMatch;
+					switches.Add(swMatch);
 				}
 			}
+			else if (!string.IsNullOrEmpty(engineSwitch.DeviceHint)) {
+				foreach (var device in switchDevices) {
+					var regex = new Regex(engineSwitch.DeviceHint, RegexOptions.IgnoreCase);
```
xantari, line 175: If you use the static instance of Regex instead of a new instance of it each time you can get a performance increase of .NET's built in RegEx caching. Additionally, if this is doing far more "lookups" then setting it to compiled is also a good idea as you'll get a perf hit the first call, but subsequent calls will be faster as it will be precompiled and cached.

So like this:
```cs 
Regex.IsMatch(device.name, engineSwitch.DeviceHint, RegexOptions.IgnoreCase | RegexOptions.Compiled)
```

See here: https://docs.microsoft.com/en-us/dotnet/standard/base-types/compilation-and-reuse-in-regular-expressions

freezy: There are a few other places as well, here it doesn't make much of a difference because it's a one-time thing and editor-only, but it might be worth optimizing elsewhere too.

freezy (approved):
Looking good, thanks a lot! :)

VisualPinball.Unity/VisualPinball.Unity/Mappings/MappingConfig.cs
xantari: Ohh didn’t realize this is editor only.

PR #344, opened 2021-11-21
Add CollisionSwitchComponent

jsm174:
This component will expose switches for the following components that inherit `IApiHittable`

This was needed as there are switches behind the drop target banks and non-slingshots in Rock.

This PR also does the following:

- updates drop target bank component to support 10 drop targets
- updates all projects to `netstandard2.1` and C# 9.0
- fixes `VisualPinball.Engine.Unity.sln` compilation errors using Visual Studio. 
- updates `FluentAssertions to 6.2.0` 
- updates `.unity` folder with latest managed dlls for compiling in Visual Studio without Unity

(Copied from `/Applications/Unity/Hub/Editor/2021.2.3f1/Unity.app/Contents/Managed/UnityEngine` and `ScriptAssemblies`)

Remaining:

- [x] Create new icons for `CollisionSwitch` (@freezy)
- [x] Support more components other than surface, primitive, and rubber? (Now supports `IApiHittable`)
- [x] Figure out why manager isn't showing enabled for some of the hits while it is registering in PinMame.  See https://github.com/freezy/VisualPinball.Engine/pull/344#issuecomment-976115294


VisualPinball.Unity/VisualPinball.Unity/VPT/CollisionSwitch/CollisionSwitchApi.cs
```diff
@@ -0,0 +1,83 @@
...
+			}
+			else if (_colliderComponent is RubberColliderComponent) {
+				_player.TableApi.Rubber(_collisionSwitchComponent.name).Hit += OnHit;
+			}
+			else if (_colliderComponent is PrimitiveColliderComponent) {
+				_player.TableApi.Primitive(_collisionSwitchComponent.name).Hit += OnHit;
+			}
+			else {
+				Logger.Error($"{_collisionSwitchComponent.name} not connected to a surface, rubber, or primitive");
+			}
```
freezy, line 56: I'd prefer if this was interfaced so we can easily add other components in the future. For the API classes there is `IApiHittable` which contains the `OnHit` events. You'll probably have to keep a hittable component dictionary in `TableApi`, but I'd prefer to do it there, because it's a more central place we're less likely to miss in the future when adding other components.

jsm174: Okay, I will make the changes tonight.

Should we rename again to HitSwitch?

freezy: I'd prefer Collision, Hit is still from the VPX code base, and I'd rather move to Collision, which seem to be the common term for that.

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableApi.cs
```diff
@@ -45,7 +45,8 @@ public class TableApi : IApi
 		private readonly Dictionary<string, TeleporterApi> _teleportersByName = new Dictionary<string, TeleporterApi>();
 		private readonly Dictionary<string, TriggerApi> _triggersByName = new Dictionary<string, TriggerApi>();
 		private readonly Dictionary<string, TroughApi> _troughsByName = new Dictionary<string, TroughApi>();
-
+		private readonly Dictionary<string, CollisionSwitchApi> _collisionSwitchesByName = new Dictionary<string, CollisionSwitchApi>();
```
freezy, line 48: Why not `IApiSwitch` instead of `CollisionSwitchApi`?

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/CollisionSwitchInspector.cs
```diff
@@ -0,0 +1,65 @@
...
+	{
+		protected override MonoBehaviour UndoTarget => throw new System.NotImplementedException();
+
+		public override void OnInspectorGUI()
+		{
+			base.OnInspectorGUI();
+
+			if (Application.isPlaying) {
+				TableApi tableApi = TableComponent.GetComponent<Player>().TableApi;
+				CollisionSwitchApi collisionSwitchApi = tableApi.CollisionSwitch(target.name);
```
freezy, line 35: Your `target` is the component, so you don't need to reference by name, no?

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableApi.cs
jsm174: Okay, so you want `switchesByName`, `switchesByComponent`? 

```
public IApiSwitch Switchable(string name) => _switchesByName[name];
public IApiSwitch Switchable(MonoBehaviour component) => _switchesByComponent[component];
```

freezy: Yeah, I think that would be more future-proof. You don't need to cast it back in the collider switch API, do you?

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/CollisionSwitchInspector.cs
freezy: Basically the name references are probably going away, since we don't rely on names anymore. I kept them because of potential GLE scripts that might use them, but internally we should use references by component wherever possible.

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableApi.cs
jsm174: Looks like I will have to make `IApiSwitch` public instead of internal.

I get inconsistent accessibility.

freezy: Yup, possible, go ahead :)

jsm174: So I tried to use `IApiSwitch` instead,  but then this would fail:

```
			var nameDict = GetNameDictionary<T>();
			var compDict = GetComponentDictionary<T>();

			nameDict[component.name] = api;
			compDict[component] = api;
```

I didn't want to put something like `if (!(api is CollisionSwitchApi) ` around it. 

Feel free to change this if you want.

VisualPinball.Unity/VisualPinball.Unity/VPT/CollisionSwitch/CollisionSwitchApi.cs
jsm174: Cool. Thx!

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/CollisionSwitchInspector.cs
jsm174: Updated, but now it looks like this:


```
CollisionSwitchApi collisionSwitchApi = (CollisionSwitchApi)tableApi.CollisionSwitch((MonoBehaviour)target);
```

Not sure if their is any more specific cast than `MonoBehaviour`

Also we could `tableApi.Switchable` but this goes back to the other comment.

jsm174:
So I had a chance to look at the switch manager status.

Screen Shot 2021-11-22 at 9 45 55 PM

Since this same switch ID is currently hooked to 6 collision switches, the status is only updated when the first device is hit.

It looks like the switch statuses are stored in a dictionary by ID. Additional switches with the same ID, just override the dictionary value. So the last component is the one that is used for the status:
 
https://github.com/freezy/VisualPinball.Engine/blob/67aca1d72b51a92e66389a46fe1ae09282d4bde3/VisualPinball.Unity/VisualPinball.Unity/Game/SwitchPlayer.cs#L87-L90

Thoughts on this?

VisualPinball.Unity/VisualPinball.Unity/VPT/Table/TableApi.cs
freezy: There already was `_switchablesByComponent` and `_switchablesByName`, using those. Needed to remove the exception because we now save references to switchables where we don't have a reference to the actual typed object.

VisualPinball.Unity/VisualPinball.Unity.Editor/Inspectors/CollisionSwitchInspector.cs
freezy:
```cs
if (switchApi is CollisionSwitchApi { IsHittable: true }) {
	DrawSwitch($"Collision switch exposed as {target.name}.", switchApi.IsSwitchEnabled);
}
```
is fine.

freezy:
> Thoughts on this?
That's okay, no? I agree it's a bit weird in the editor, but any of those elements triggers the same switch, so it's kinda logical that they all change status synchronously.

The other way of handling this would be to additionally store switch statuses in a switch mapping -> status dictionary, just for the editor. I'd create a GitHub issue about this and move along. :)

freezy (approved):
For me it's good to merge!

jsm174:
> That's okay, no? I agree it's a bit weird in the editor, but any of those elements triggers the same switch, so it's kinda logical that they all change status synchronously.

Sorry, I guess I didn't explain it properly.  I'll create another issue.

Switch `41` is currently used in 8 places, for example:

- `A` (Behind upper drop target bank)
- `B` (Behind lower drop target bank)
- `C` (Upper left side rubber)
- `D` (Upper right side rubber)
- `E` (Middle left side rubber)
- `F` (Middle right side rubber)
- `G` (Left slingshot - FUTURE)
- `H` (Right slingshot - FUTURE)

If the user hits `A`, all `41` entries in the Switch Manager close. Totally fine. 

Screen Shot 2021-11-23 at 6 51 53 PM
 
However, if the user hits `B`, `C` , `D`, `E`, `F`, `G`, `H`, the entries in the switch manager will not close.

The Game Logic Engine will work fine and PinMame registers the switch for `A` - `H`.

The SwitchPlayer statuses are a dictionary by ID `41`
 https://github.com/freezy/VisualPinball.Engine/blob/67aca1d72b51a92e66389a46fe1ae09282d4bde3/VisualPinball.Unity/VisualPinball.Unity/Game/SwitchPlayer.cs#L87-L90

So when the player is populating the dictionary, each ID `41` switch device overwrites the last one.  

So that's why the manager looks like its only monitoring one device.

PR #346, opened 2021-11-25
Add Slingshot API and expose slingshot switch

jsm174:
The PR adds switches to slingshots. 

The switch will be enable when a `Slingshot` event occurs

Screen Shot 2021-11-24 at 7 04 48 PM
.

Remaining:

- [ ] ?


freezy (approved):
Awesome, thanks a lot! :)

Will now add bumpers as well.

PR #353, opened 2021-12-14
Fix editor performance 

jsm174:
Currently, the Lamp/Switch/Coil/Wire managers are refreshed during gameplay using:

```cs
#if UNITY_EDITOR
   UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
#endif
```

@Pandelii noticed a huge performance hit and after debugging, this was the culprit.

This PR adds a new `Update During Gameplay`setting to the `Player` component that is defaulted true.
When this is enabled and a switch, coil, or lamp is updated, we will look for any open corresponding managers, and call the `repaint` method. (note, there can be multiple of the same manager windows open)

We also do the same for the `TroughInspector` and `CoilDevice`.

This PR also does the following:

- cleans up the lamp groups for Rock in preparation for Visual Scripting 
- updates `NetVips`, `NetVips Native`, and `NLog` dependencies to the latest. 
- reverts the hacked in Lamp Type of `Single Off On` in favor of upcoming Visual Scripting.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/Game/CoilPlayer.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/Game/CoilPlayer.cs
@@ -44,6 +44,7 @@ namespace VisualPinball.Unity
 		private IGamelogicEngine _gamelogicEngine;
 		private LampPlayer _lampPlayer;
 		private WirePlayer _wirePlayer;
+		private Player _player;
 
 		private bool _updateDuringGameplay = false;
 
@@ -62,7 +63,7 @@ namespace VisualPinball.Unity
 			_lampPlayer = lampPlayer;
 			_wirePlayer = wirePlayer;
 
-			_updateDuringGameplay = UnityEngine.Object.FindObjectOfType<Player>().UpdateDuringGamplay;
+			_player = UnityEngine.Object.FindObjectOfType<Player>();
 		}
 
 		public void OnStart()
@@ -164,9 +165,7 @@ namespace VisualPinball.Unity
 				}
 
 #if UNITY_EDITOR
-				if (_updateDuringGameplay) {
-					RepaintManagers();
-				}
+				RefreshUI();
 #endif
 
 			} else {
@@ -182,8 +181,12 @@ namespace VisualPinball.Unity
 		}
 
 #if UNITY_EDITOR
-		private void RepaintManagers()
+		private void RefreshUI()
 		{
+			if (!_player.UpdateDuringGamplay) {
+				return;
+			}
+
 			foreach (var manager in (EditorWindow[])Resources.FindObjectsOfTypeAll(Type.GetType("VisualPinball.Unity.Editor.CoilManager, VisualPinball.Unity.Editor"))) {
 				manager.Repaint();
 			}
--- a/VisualPinball.Unity/VisualPinball.Unity/Game/CoilPlayer.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/Game/CoilPlayer.cs
@@ -14,9 +14,17 @@
 // You should have received a copy of the GNU General Public License
 // along with this program. If not, see <https://www.gnu.org/licenses/>.
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
+using UnityEngine;
 using NLog;
+using Logger = NLog.Logger;
+
+#if UNITY_EDITOR
+using UnityEditor;
+
+#endif
 
 namespace VisualPinball.Unity
 {
@@ -37,6 +45,8 @@ namespace VisualPinball.Unity
 		private LampPlayer _lampPlayer;
 		private WirePlayer _wirePlayer;
 
+		private bool _updateDuringGameplay = false;
+
 		private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
 
 		internal Dictionary<string, bool> CoilStatuses { get; } = new Dictionary<string, bool>();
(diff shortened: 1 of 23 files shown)


freezy:
image

Niiice! Thanks for that!

freezy:
Would it be easy to make this toggleable during gameplay? Right now it seems you need to stop, toggle, and start, to get the effect.

VisualPinball.Unity/VisualPinball.Unity/Game/Player.cs
```diff
@@ -51,6 +51,9 @@ public class Player : MonoBehaviour
 		[HideInInspector] [SerializeField] public string debugUiId;
 		[HideInInspector] [SerializeField] public string physicsEngineId;
 
+		[Tooltip("UpdateDuringGamplay")]
```
freezy, line 54: Maybe something like "When enabled, update the switch, coil, lamp and wire manager windows in the editor (slower performance)"?

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
```diff
@@ -282,7 +286,7 @@ private void HandleKeyInput(object obj, InputActionChange change)
 									}
 								}
 								#if UNITY_EDITOR
-									UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
+									//UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
```
freezy, line 289: Still TODO?

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
```diff
@@ -310,7 +314,7 @@ public void HandleSwitchChange(WireDestConfig wireConfig, bool isEnabled)
 								wire.OnChange(false);
 								WireStatuses[wireConfig.Id] = (false, 0);
 								#if UNITY_EDITOR
-									UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
+									//UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
```
freezy, line 317: Still TODO?

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
```diff
@@ -328,7 +332,7 @@ public void HandleSwitchChange(WireDestConfig wireConfig, bool isEnabled)
 									wire.OnChange(false);
 									WireStatuses[wireConfig.Id] = (false, -2);
 									#if UNITY_EDITOR
-										UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
+										//UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
```
freezy, line 335: Still TODO?

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
```diff
@@ -338,7 +342,7 @@ public void HandleSwitchChange(WireDestConfig wireConfig, bool isEnabled)
 						}
 					}
 					#if UNITY_EDITOR
-						UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
+						//UnityEditorInternal.InternalEditorUtility.RepaintAllViews();
```
freezy, line 345: Still TODO?

freezy (approved):

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
jsm174: Fixed. I don't know how I forgot these.

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
jsm174: Fixed. I don't know how I forgot these.

VisualPinball.Unity/VisualPinball.Unity/Game/WirePlayer.cs
jsm174: Fixed. I don't know how I forgot these.

VisualPinball.Unity/VisualPinball.Unity/Game/Player.cs
jsm174: Updated

jsm174:
> Would it be easy to make this toggleable during gameplay? Right now it seems you need to stop, toggle, and start, to get the effect.

Cool. I renamed all the `Repaint*` methods to `RefreshUI` and put a check inside for `_player.UpdateDuringGameplay`. This allow the flag to be toggled during gameplay.

VisualPinball.Unity/VisualPinball.Unity/Game/CoilPlayer.cs
```diff
@@ -51,6 +59,8 @@ public void Awake(TableComponent tableComponent, IGamelogicEngine gamelogicEngin
 			_gamelogicEngine = gamelogicEngine;
 			_lampPlayer = lampPlayer;
 			_wirePlayer = wirePlayer;
+
+			_player = UnityEngine.Object.FindObjectOfType<Player>();
```
freezy, line 63: I'd prefer passing it through the constructor like the other references, instead of globally fetching it. I know there are a few other places where we do that, but the problem with that approach is that if for whatever reason multiple player components need to be managed, this will break.

jsm174: Fixed! 

The only place left is inside the `DeviceCoil`. Not sure what to do here.

freezy: Thanks! I'll have a look at `DeviceCoil`.

freezy: Pushed an update. Passing it through the constructor as well, since everywhere it's instantiated we have access to the `Player` instance.

jsm174: awesome. Thank you!

PR #356, opened 2021-12-21
Update index.md

DanLShane:
Added a couple of commas.

Question: Is the word "normals" previously defined in the documentation?


freezy (approved):
Thanks!

PR #359, opened 2021-12-22
Update 2-create-mesh.md

DanLShane:
I am curious about the significance of "numpad ." in line 15.  I am guessing it's some sort of key-combo, but I'm not sure if it needs correction.


freezy (approved):
Thanks!

PR #365, opened 2022-01-06
Add support for Unity Visual Scripting

jsm174:
This PR adds an API to existing GLEs so they can better interact with other components, namely the Visual Scripting package.

In addition to above, the follow other changes have been made:

- bump to latest unity deps
- add new patcher for Volley
- updates light groups in Rock patcher
- lamp values are now always from 0-255 due to a change in PinMAME.
- lamp groups can now contain other lamp groups.


VisualPinball.Unity/VisualPinball.Unity/VisualPinball.Unity.asmdef
```diff
@@ -11,6 +11,7 @@
 		"Unity.Physics.Hybrid",
 		"Unity.Transforms",
 		"Unity.InputSystem",
+		"Unity.VisualScripting.Core",
```
freezy, line 14: Can remove it from here as well

freezy (approved):
Thanks!

PR #368, opened 2022-01-14
Feature/mwg2

Cupiii:
Finished the metal wire guide. 
- Thickness (as float) and Standheight added
- Better default values
- Fixed Normals causing physics issues
- unit tests

Also fixed the Ball not rotating properly (as far as i can test).


Cupiii:
So this time, I tested before PR, don't know what's up with the docker image not testing inside unity.

freezy:
Yeah I think the Unity License we use at GitHub doesn't work for forks.

Anyway, thanks a lot for this! The ball rotation is much better, but still not perfect. I'm not sure whether now it's a problem that is apparent in VPX as well. 

This is the ball rotation on current master:

https://user-images.githubusercontent.com/70426/149617680-1ddb3a4b-fac8-450d-90cd-b1c56f8458b6.mp4

And here the ball rotation of this PR:

https://user-images.githubusercontent.com/70426/149617688-833d9703-5b77-4d33-9a11-580f62f4d9f4.mp4

It's actually a bit deceptive because the PR ball rotation is so much better in general, yet here it seems like master ball rotation is more natural. But it's a good and easily reproducible example.

You can turn on slow-motion by pressing the `S` key. Let me know if you want to take another shot at it, otherwise I'll merge since this is much better already.

Cheers and thanks again!

freezy (approved):

Cupiii:
Yeah, I know it’s not perfect yet, but actually it’s somehow complicated (maybe only for me) to test. But i wont change it again in the near future. Maybe I‘m doing something wrong when trying to test. Testing is VERY unreliable for me. Maybe i need another session with you, freezy. I will PM you.

freezy:
Awesome, merging.

PR #373, opened 2022-01-22
Add support for On Screen Input System buttons

jsm174:
Unity's Input System has support for On Screen Controls.  On Screen Controls can simulate device input. For example, an `OnScreenButton` can trigger a keypress. This is great, but you have to specify which key to listen for.

VPE has Input Actions defined such as `Left Flipper`. To make it easier to associate an `OnScreenButton` with `Left Flipper`, we added a new `OnScreenInputSystemButton` class. It extends `OnScreenButton` and uses the first binding found for the specified Action Map / Action.

Here is a video:

https://user-images.githubusercontent.com/1197137/150619707-0b94b843-fe19-48ce-a9b9-5268306e1aee.mp4

In addition, this PR updates the following:

- changes the `OnPlayerStarted` event  to `OnStarted` in Gamelogic Engine.  For PinMame for example, we can not trigger `OnStarted` until the game actually starts. 
- updates `CameraTranslateAndOrbit.cs` to support touch (pan / pinch / zoom)
- updates the test project to the latest unity and dependencies
- add `ugui` to dependencies to support `OnScreenControl`
- adds a constant in `InputManager` for `Assets/Resource` since this it is used in multiple places  

PRs for VPE.PinMame, and VPE.VisualScripting will follow and can only be merged after this PR.


freezy (approved):
Thanks, looking good. Are you planning on writing documentation too, or is this still too bleeding edge?

jsm174:
I can work on some docs. I don't know if it's bleeding edge. We don't really know how we are going to do a UI yet.

This can be more of a debug thing when testing on mobile devices.

PR #375, opened 2022-01-28
DropTargets: QUICKFIX - fix Sluggish ball when rolling over dropped DT

Cupiii:
Does not fix triggering more than once when ball in hand and possibly even in normal play


VisualPinball.Unity/VisualPinball.Unity/Physics/Collision/StaticNarrowPhaseSystem.cs
```diff
@@ -124,16 +127,44 @@ protected override void OnUpdate()
...
+								case ColliderType.Circle:
+								case ColliderType.LineZ:
+								case ColliderType.Plane:
+								case ColliderType.Point:
+								case ColliderType.Triangle:
+									// hit target
+									if (coll.Header.ItemType == ItemType.HitTarget) {
+										var normal = coll.Type == ColliderType.Triangle
+											? ((TriangleCollider*)collider)->Normal()
+											: collEvent.HitNormal;
```
freezy, line 142: `normal` isn't used, right? Can I remove it? (I'm cleaning up a few other things, too.)

VisualPinball.Unity/VisualPinball.Unity/Physics/Collision/StaticNarrowPhaseSystem.cs
```diff
@@ -124,16 +127,44 @@ protected override void OnUpdate()
...
+										if (coll.Header.IsEnabled) { }
```
freezy, line 143: Also, this does nothing, good to remove?

freezy (approved):
Tested, and works great. Lemme know about the two comments, then I push my changes and merge.

Thanks! :)

VisualPinball.Unity/VisualPinball.Unity/Physics/Collision/StaticNarrowPhaseSystem.cs
Cupiii: yeah, screw it.

VisualPinball.Unity/VisualPinball.Unity/Physics/Collision/StaticNarrowPhaseSystem.cs
Cupiii: yup, you also can remove that.

freezy:
All in. Cheers! :)

Issue #380, opened 2022-02-19
The BallRingCounterSystem is not required anymore


simonbethke:
I reversed a bit what the BallSpinHack is and what it is used for:

The BallSpinHack is a system which checks old positions in order to find if the ball is still moving. If it is not moving anymore the angular momentum is dampened. For that it utilizes the BallRingCounter and this ball ring counter even has an own system. 

The BallRingCounter has been [introduced in VPX for ball-trails](https://github.com/vpinball/vpinball/blob/be08b04d61096272df97bd45e6f0682043228a73/pin/player.cpp#L5463). As ball-trails are either not required anymore or if required are done by unity stuff this system is not required anymore. And actually the trails itself have already been removed.

Of course VPX grew not only historically but also hysterically. In this case that means that the positions for the trails have simply been repurposed for the spinhacksystem. 

As the spinhacksystem is 'only' used to apply playfield friction to the ball I suggest to remove both systems and add the dampening back in later on when implementing playfield friction.

PR #381, opened 2022-02-19
VPE-380 Removing BallRingCounterSystem and BallSpinHackSystem

simonbethke:
see https://github.com/freezy/VisualPinball.Engine/issues/380


Cupiii:
Thanks for pointing out the the ring-counter is only needed for the ballspinhack and possibly not needed in future anymore. 

The question is, what do we loose here, when we delete all this? 

We currently have problems with exessive ball spin in VPE and don't know yet, where it comes from (at least I have and I don't know). Especially when hitting rubbers relatively perpendicular to the normal. We also have ball spin issues on the flippers, I believe. 
Also we had problems only a little time ago with restless balls on flippers and other objects. Especially rubbers but also walls and MWG. 

I don't know if the Ballspinhack reduces current issues or maybe even is the culprit here. For Rubbers (and mwg) I have the Idea, that the colliders may be a little bit too pointed, but have not yet tested. That usually is the problem that causes such issues, as I read...

@simonbethke Did you also experience the ballspin issues before removing the hack and does it differ after?

Also, @simonbethke , you mention that the ballspinhack would not be needed, if something is programmed in the future. Wouldn't it be right to program it and then delete the hack, when the implementation works (and just comment it out, meanwhile?) , especially when we currently have problems which the hack "should" solve (excessive spin and rotation while static (restless balls))?

Issue #377, opened 2022-01-29
Rubbers do not colllide sometimes


Cupiii:
After some playtest, I think that Rubbers do not collide either at very tight position with many splinepoints or at plinepoint 0 and its neighbor. 

Sometimes the ball flies through. 

While working on the metalwireguides I saw the meshes twisting somehow. 

With the MWG-code (and some intelligent copy and paste) it should be fairly easy to mitigate or fix the problem, if I'm correct, that the twisting meshes are the culprit here.

Cupiii:
After Commits on 13.th Mar, the problem got better, but is not yet fixed fully. Either we have to set some walls behind the rubbers or add an option to do that while creating the rubber.

PR #384, opened 2022-03-12
Rubbers meshGen fix and Gizmo positioning fix for rubbers and MWG

Cupiii:
Fixes Issue "Rubbers do not colllide sometimes #377".

Wrote Rubbers from scratch (copied some code from MWG).

Also fixed UV Mapping


freezy:
Cool! Is this something that could go back to VPX?

freezy (approved):
Very cool! There is a warning about `_middlePoint`, says it's never assigned but used. Can it be removed?

Cupiii:
Have to look at it tomorrow.

Cupiii:
_middlepoint is set again for rubbers and MWG. Test run fine on my computer.

Cupiii:
Oh, and yes, that may be backported to VP as well, but check for MWG branch, where I edited something in SplineVertex class. I think I only made the 2d spline non loopable, but i may have altered more without knowing. I should have used the 3dSpline Class which was used for Ramps... but did not...

PR #386, opened 2022-03-13
Ball orientation Fix (and documentation)

Cupiii:
Ball orientation seems to be fine now. 

Still strange things happen on rubbers and on flippers. But slow ball movement visualized correctly now.


freezy (approved):
Too good to be true! :)

Thanks for this.

PR #393, opened 2022-03-21
Ball-orientation-Nan-fix

Cupiii:
This fixes the ball orientation lock that occoured after a while playing.

The Skew Matrix produced more and more longer vectors in the orientation matrix. 
Fix is to "kind of" normalize the single vectors inside the orientation matrix. 
"Kind of": we don't need to normalize fully (dividing by sqrt(x^2+y^2+z^2)), since we don't use the vector lengthes anywhere (and also should not use them anywhere).
So we can divide by the sum of absolute values of x, y and z). Giving "somewhat" normalized vectors.

Also made the BallMovementsystem a little bit more clear, introducing some inbetween-vectors. I think the clearity is worth the small performance cost.

While analysing I added some comments about general ball-physics that may be useful in the future.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
@@ -26,7 +26,24 @@ namespace VisualPinball.Unity
 		public float3 Position;
 		public float3 EventPosition; // m_lastEventPos
 		public float3 Velocity;
+
+		/// <summary>
+		/// AngularVelocity 
+		///		* Set to 0 at start
+		///			(in BallManualRoll(in Entity entity, in float3 targetWorldPosition)
+		///			(which is not used anywhere)
+		///		* Used to get  tangential velocity due to rotation when rolling / colliding on surfaces (alsways added to normal velocity) 
+		///			(in BallData.SurfaceVelocity(in BallData ball, in float3 surfP))
+		///		* Calculated from AngularMomentum / inertia 
+		///			(In BallDisplacementSystem.OnUpdate())
+		///			(Where Inertia is a "constant" based on radius and mass (2/5 m r^2))
+		/// </summary>
 		public float3 AngularVelocity;
+		/// <summary>
+		/// AngularMomentum
+		///		* set to 0 at Start.
+		///		* When a survace applies an impule, it applies it to velocity and to angMom. ()
+		/// </summary>
 		public float3 AngularMomentum;
 		public float3x3 Orientation;
 		public float Radius;
@@ -72,7 +89,12 @@ namespace VisualPinball.Unity
 			}
 		}
 
+		/// <summary>
+		/// Calculates Moment of Inertia for a Ball 
+		/// https://en.wikipedia.org/wiki/Moment_of_inertia#Examples_2
+		/// </summary>
 		public float Inertia => 2.0f / 5.0f * Radius * Radius * Mass;
+		
 		public float InvMass => 1f / Mass;
 
 		public void ApplySurfaceImpulse(in float3 rotI, in float3 impulse)
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
@@ -29,20 +29,39 @@ namespace VisualPinball.Unity
 
 		/// <summary>
 		/// AngularVelocity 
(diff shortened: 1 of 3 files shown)


VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
```diff
@@ -26,7 +26,43 @@ internal struct BallData : IComponentData
 		public float3 Position;
 		public float3 EventPosition; // m_lastEventPos
 		public float3 Velocity;
+
+		/// <summary>
+		/// AngularVelocity  -  german: Winkelgeschwindigkeit
+
+		///		* Set to 0 at Manual Roll
+		///			(in BallManualRoll(in Entity entity, in float3 targetWorldPosition)
+		///			(which is not used anywhere)
```
freezy, line 35: It's used [here](https://github.com/ravarcade/VisualPinball.Engine.Unity.ImgGUI/blob/master/Monitors/BallMonitor.cs#L36), but we can probably remove it unless we want to make the physics API agnostic again.

VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallDisplacementSystem.cs
```diff
@@ -57,6 +57,37 @@ protected override void OnUpdate()
 				ball.Orientation += addedOrientation;
 				math.orthonormalize(ball.Orientation);
 
+				// after Othonomalization, the Orientation vectors also have to be normalized - this is not done by othonomalize, since the skew matrix creates quite lengthy vectors.
```
freezy, line 60: Othonomalization? Orthonormalization? :)

VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallDisplacementSystem.cs
```diff
@@ -57,6 +57,37 @@ protected override void OnUpdate()
 				ball.Orientation += addedOrientation;
 				math.orthonormalize(ball.Orientation);
 
+				// after Othonomalization, the Orientation vectors also have to be normalized - this is not done by othonomalize, since the skew matrix creates quite lengthy vectors.
+				// in fact, they dont have to be normalized, but just shortened, so we can abs-add the x, y and z together and just divide by the sum.
+				// This saves three qrts in the game loop per ball.
```
freezy, line 62: What's a qrt?

freezy (approved):
Awesome. I'm wondering how you came across the orthonormalization thing. But it looks solid.

The only question I have is why is it that VP doesn't seem to have this NaN issue? Is there code fixing the symptom that I've missed when porting?

VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
Cupiii: Oh, OK. did not see this and it was not signed as a "reference" inside of Visual Studio. Will add that. I wondered why it was not used. Now I know that I have to search for the Function name across the repo. Thanks for pointing out, will add that.

VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallDisplacementSystem.cs
Cupiii: typo...

VisualPinball.Unity/VisualPinball.Unity/VPT/Ball/BallData.cs
Cupiii: Oh, now i see, it's a Plugin - a physics debugger... had i known about this... Seems nice

Cupiii:
Actually I debugged the vector and the values grew and grew until the two vectors used in the balltransform.localrotation were almost the same (maybe because of rounding errors with higher values) That gave assertion errors in the localrotation method.  I just looked at the self-written VP-Orthonomalization code and it seems that VP does a Orthonomalization (including a normalization) there, but I'm actually not sure. Usually Orthonomalisation does not "normalize" the vectors length wise. So if VP does it within orthonormalize, it's at least unusual. 

So I took a look:
VP's code is here: https://github.com/vpinball/vpinball/blob/be08b04d61096272df97bd45e6f0682043228a73/math/matrix.h#L208
VP actually does not do a true Orthonormalisation...  and normalizes while orthonomalizing.

Maybe VP is faster - gave hint in the code for optimization.

PR #395, opened 2022-03-27
Feature: curve based dampeners2

Cupiii:
CBD aka nFozzy rubber dampeners are ready for now.,. although they are not exactly nFozzys rubber dampeners. Beside the elasticity based effects, nFozzy also dampens the frictions based effects. That may be intentional, but it may also be only because he could not get "inbetween" these calculations. Anf from my research (and playtest (see below)) different curves for both effects seem to be beneficial for the "feel" of the game.
I'm with RothbauerW in (private, for now) contact to discuss the differences. Also a book about contact physics is arriving at my place on tuesday.
We may also open the discussion in the physics thread on discord.

What's great: nFozzy physics now can be set for every collider seperately. CBD on MWG are a must. 
I dont know if this is right, but I really think that different CBD on each material will set new standards to Virtual Pinball. 

As said above: There is a new (experimental) feature beside nFozzy: friction over velocity. After a short playtest this feels really great. especially when you stop the ball with the flipper pressed and it rolls to the metal. Rotation looks much, much more natural now and feels great. 

This is a first look, not meanted to be pulled now. If anybody has suggestions...


Cupiii:
Oh, although I‘m wasting some memory with this (1kb per collider), i see no impact in FPS on my computer.

Cupiii:
OK, ready to merge.

VisualPinball.Unity/VisualPinball.Unity/VPT/PhysicsMaterialComponent.cs
```diff
@@ -1,38 +1,80 @@
...
+	/// <summary>
+	/// A physical material used by the physics engine<p/>
+	///
+	/// Materials are actually a big deal in VP, authors as well as players
+	/// tweak them all the time, so getting those from external assets instead
+	/// of writing them into the scene seems a good plan.
+	/// </summary>
+	[CreateAssetMenu(fileName = "PhysicsMaterial", menuName = "Visual Pinball/Physics Material", order = 100)]
+	//[CustomEditor(typeof(PhysicsMaterial))]
+	public class PhysicsMaterialComponent : ScriptableObject
```
freezy, line 34: Any particular reason to rename this? It's not a component, so it's a bit confusing. If you wanna be more precise, I'd propose `PhysicsMaterialAsset`, which is how scriptable objects are stored.

VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/PhysicsMaterialInspector.cs
```diff
@@ -0,0 +1,217 @@
...
+
+
+		private void OnEnable()
+		{
+			Elasticity = serializedObject.FindProperty("Elasticity");
+			ElasticityFalloff = serializedObject.FindProperty("ElasticityFalloff");
+			ElasticityOverVelocity = serializedObject.FindProperty("ElasticityOverVelocity");
+			Friction = serializedObject.FindProperty("Friction");
+			ScatterAngle = serializedObject.FindProperty("ScatterAngle");
+			FrictionOverVelocity = serializedObject.FindProperty("FrictionOverVelocity");
```
freezy, line 55: I would prefer using `nameof(PhysicsMaterialComponent.Elasticity)`, which is more refactor-friendly.

freezy (commented):
Geil! :)

Very cool. Two things:

- At some point it would be nice to include a bunch of materials made for typically used components, and assign them to the bunch of prefabs we already have.
- Documentation would be nice. Specially how it differs from nFozzy, how to achieve which effect, etc. I can setup a page if you want.

VisualPinball.Unity/VisualPinball.Unity/VPT/PhysicsMaterialComponent.cs
Cupiii: *Asset would be good. My problem was that we had a struct and the asset, named the same. Renaming will break the current table again though, but can do this tomorrow.

VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/PhysicsMaterialInspector.cs
Cupiii: Oh, yes. did not know that this works. Great.

Cupiii:
All resolved. Could be pulled now. It will destroy the V-Table, I will correct it tomorrow then.

PR #404, opened 2022-04-10
Feature-FlipperTricks

Cupiii:
This adds FlipperTricks by RothbauerW. 
I implemented this exactly like RothBauerW did in his script (same variable names). 
There is no microbounce to the ball yet, if releases an ball is in cradle - this will come together with dropcatch.

To test, please also switch branch at the table.

FlipperTricks is off by default (on new Flippers).

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/ItemInspector.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/ItemInspector.cs
@@ -147,9 +147,16 @@ namespace VisualPinball.Unity.Editor
 			EditorGUI.BeginChangeCheck();
 
 			if (string.IsNullOrEmpty(label)) {
-				EditorGUILayout.PropertyField(serializedProperty);
+				if (string.IsNullOrEmpty(serializedProperty.tooltip))
+					EditorGUILayout.PropertyField(serializedProperty);
+				else
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(serializedProperty.displayName, serializedProperty.tooltip));
 			} else {
-				EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label));
+				if (string.IsNullOrEmpty(serializedProperty.tooltip))
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label));
+				else
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label, serializedProperty.tooltip));
+
 			}
 
 			if (EditorGUI.EndChangeCheck()) {
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperStaticData.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperStaticData.cs
@@ -38,7 +38,10 @@ namespace VisualPinball.Unity
 		public float OriginalTorqueDamping;
 		public float OriginalTorqueDampingAngle;
 		public float ElasticityMultiplier;
-		public bool lastSolState;
+		public bool LastSolState;
+		public bool WasInContact;
+		public bool UseFlipperTricks;
+
 		// externals
 		public float SOSRampUp;
 		public float SOSEM;
@@ -53,5 +56,7 @@ namespace VisualPinball.Unity
 		public float EndRadius;
 		public float FlipperRadius;
 
+		// for debugging while in burst
+		//public float ft;
 	}
 }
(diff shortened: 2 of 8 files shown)


VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperStaticData.cs
```diff
@@ -31,9 +31,32 @@ internal struct FlipperStaticData : IComponentData
...
+
+		// externals
+		public float SOSRampUp;
+		public float SOSEM;
+		public float EOSReturn;
+		public float EOSTNew;
+		public float EOSANew;
+		public float EOSRampup;
+		public float Overshoot;
+		public float BumpOnRelease;
```
freezy, line 53: I think it's better to put those into another component class (maybe `FlipperTrickData`?). The reason is that `FlipperStaticData` is used in two systems (`FlipperDisplacementSystem` and `FlipperVelocitySystem`), but the data you added is only accessed in `FlipperVelocitySystem`. In general, the idea of DOTS is that you split your data into pieces as small as possible, to never load data that isn't needed.

freezy (commented):
Very cool! We need a SS table to test all this :)

Changelog and some doc would be nice before merging. Other comments below.

VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/ItemInspector.cs
```diff
@@ -147,9 +147,16 @@ protected void EndEditing()
...
+					EditorGUILayout.PropertyField(serializedProperty);
+				else
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(serializedProperty.displayName, serializedProperty.tooltip));
 			} else {
-				EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label));
+				if (string.IsNullOrEmpty(serializedProperty.tooltip))
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label));
+				else
+					EditorGUILayout.PropertyField(serializedProperty, new GUIContent(label, serializedProperty.tooltip));
+
```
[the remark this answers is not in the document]
Cupiii: yep, noticed they don't work.

VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperStaticData.cs
Cupiii: Do you really believe that it impacts performance positively to add another iComponentData? I actually thought about it and came to the conclusion, that adding it (50 Bytes) to the existing data would be better performance wise, since chunks would be almost empty with only 2 to 4 flippers at a table, and with only 50 Bytes more, the cache misses would not increase so much, I believe, considering chaches in Kilobyte ranges. Instead, I thought, that adding another iComponentData would add much overhead with only two to four instances/entities.  On the other hand, i had to declare the existing data writable (which could be problematic for parallelisation), but also - with only a few entities, i doubt that we might see better performance with a new extra iComponentData. 
Would love to hear your opinion. (actually i could find arguments for both opinions online, and it resulted in trial and error, but almos always related to many more entities in a system, where size relly does matter for keeping all in cache)
Even for Readability I'm not sure what would better... But "FlipperStaticData" should be renamed to "FlipperData", if we keep current implementation.

freezy: You're probably right that the performance impact is minimal to none. It's just how DOTS is designed to work, and the difference is surely measurable when thousands of entities are in play. Which will never be the case here.

The only other argument I can think of is code clarity. It might be better to understand if the flipper-tricks related code also uses flipper-tricks related data. And not write something that is called `Static` (renaming it to `FlipperData` only half-solves the problem, since there actually *is* quite some static data).

I'll let you decide. I would prefer two classes, but I'll accept the PR if you think otherwise.

Cupiii: OK, thanks. I will think a little bit about this while preparing for the live catch and flipper-roll-bumping. I would need some more data there, i guess. I don't knwo yet how to "circumvent" the parallelism when implementing that. Lets see, if i have to do a shared component for this... This would speak for a dedicated class for all those "tricks". 👍

freezy:
BTW, cc @rothbauerw

xantari:
How come this isn't merged yet?

Cupiii:
> How come this isn't merged yet?

New "Flipper_Tricks_2" (without using FlipperStaticData) PR is on the way. This PR will be closed.

Cupiii:
Will delete branch later.

PR #407, opened 2022-04-12
ci: update osx to support dual binary builds. bump to unity 2021.3.0f1

jsm174:
After trying to make a build of Rock that target `osx-arm64` and `osx-x64`, we ran into some issues, see [here](https://forum.unity.com/threads/apple-silicon-intel-64-bit-managed-and-native-library-settings-not-working-as-expected.1263548/#post-8026151).

The solution is to use dual architecture binaries. 

We created dual architecture binaries for `libpinmame`, see [here](https://github.com/vpinball/pinmame/pull/57).
We also created dual architecture binaries for `netminiz`, see [here](https://github.com/jsm174/net-miniz/commit/9b7843736b1a932e487dd1a4529c6de96bbf9c8c).

Note: NetVips does not currently have dual architecture binaries. We could use lipo on them, but then we would have to make the `.csproj` that executes commands only for osx, and that doesn't seem right at this time -- and since NetVips isn't used for Standalone builds, this isn't an issue. 

This PR refactors the `osx-arm64` / `osx-x64` plugins folders to `osx`, `osx/arm64`, and `osx/x64`. All `*.meta` files have been updated accordingly. 

This PR also updates some outdated dependencies, and updates Unity to `2021.3.0f1`.


jsm174:
Also, after this PR is merged, we will update `VisualPinball.Engine.PinMAME` to reference the latest VPE.

freezy (approved):
Cool, thanks a lot!

I'll let you merge. :)

PR #408, opened 2022-04-18
Remove internal id from switches, lamps, and coils

jsm174:
Switch, lamps, and coils in VisualPinball.Engine are identified with an `Id` variable that is a string.

PinMAME internally uses integers to identify switches, lamps, and coils.

Because of this, an `InternalId` variable was added to VisualPinball.Engine to help mapping.

When working on RGB lamps for TWD, we had issues and needed to pass `InternalId` to the lamp player. 

We then noticed missing lamps when auto `Populating` lamps in the lamp manager. Lamps that don't have an internal id default as 0. Auto `populating` overwrites existing entries by internal ids. 

Since `InternalId` is real only needed for PinMAME, (Visual Scripting and MPF only use `Id`), we could remove `InternalId` all together, and make the PinMAME GLE deal with the `Id`.

This PR does exactly that. There are additional PRs for PinMAME, Visual Scripting, and MPF.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
@@ -136,8 +136,6 @@ namespace VisualPinball.Unity.Editor
 			}
 			cellRect.x += 20;
 			cellRect.width -= 20;
-
-			EditorGUI.LabelField(cellRect, lampListData.InternalId.ToString());
 		}
 
 		protected override void RenderDeviceElement(LampListData listData, Rect cellRect, Action<LampListData> updateAction)
--- a/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
@@ -134,8 +134,6 @@ namespace VisualPinball.Unity.Editor
 				EditorGUI.DrawTextureTransparent(iconRect, icon, ScaleMode.ScaleToFit);
 				GUI.color = guiColor;
 			}
-			cellRect.x += 20;
-			cellRect.width -= 20;
 		}
 
 		protected override void RenderDeviceElement(LampListData listData, Rect cellRect, Action<LampListData> updateAction)
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Text.RegularExpressions;
 using NLog;
 using UnityEngine;
 using VisualPinball.Engine.VPT;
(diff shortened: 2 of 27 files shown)


VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
```diff
@@ -136,8 +136,6 @@ private void RenderCoilId(Dictionary<string, LampState> lampStatuses, LampListDa
 			}
 			cellRect.x += 20;
 			cellRect.width -= 20;
-
-			EditorGUI.LabelField(cellRect, lampListData.InternalId.ToString());
```
freezy, line 140: I think you can remove the cellRect part as well.

VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
```diff
@@ -196,10 +197,14 @@ void IApi.OnInit(BallManager ballManager)
...
-						_stackSwitches[sw.InternalId - 1] = CreateSwitch(sw.Id, false, MainComponent.Type == TroughType.ModernOpto ? SwitchDefault.NormallyClosed : SwitchDefault.NormallyOpen);
-						_switchLookup[sw.Id] = _stackSwitches[sw.InternalId - 1];
+				foreach (var @switch in MainComponent.AvailableSwitches) {
+					var match = new Regex(@"^ball_switch_(\d+)$").Match(@switch.Id);
+					if (match.Success) {
+						int.TryParse(match.Groups[1].Value, out int id);
+						if (id > 0) {
+							_stackSwitches[id - 1] = CreateSwitch(@switch.Id, false, MainComponent.Type == TroughType.ModernOpto ? SwitchDefault.NormallyClosed : SwitchDefault.NormallyOpen);
+							_switchLookup[@switch.Id] = _stackSwitches[id - 1];
+						}
```
freezy, line 207: This is a bit ugly, but since it's only done once I guess there isn't too much of an impact. I would at least move the regex compilation out of the loop though.

And one thing I would add is a comment when `ball_switch_{i}` is defined, so it's clear we cannot rename it without changing the regex.

VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughComponent.cs
```diff
@@ -208,7 +208,7 @@ public override TroughData CopyDataTo(TroughData data, string[] materialNames, s
 					case TroughType.ModernOpto:
 					case TroughType.ModernMech:
 						return Enumerable.Repeat(0, SwitchCount)
-							.Select((_, i) => new GamelogicEngineSwitch($"ball_switch_{i + 1}", i + 1)
+							.Select((_, i) => new GamelogicEngineSwitch($"ball_switch_{i + 1}")
```
freezy, line 211: Also add a comment so it's clear this is used in the regex.

VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughComponent.cs
```diff
@@ -219,7 +219,7 @@ public override TroughData CopyDataTo(TroughData data, string[] materialNames, s
 						return new[] {
 							new GamelogicEngineSwitch(EntrySwitchId) { Description = "Entry Switch" }
 						}.Concat(Enumerable.Repeat(0, SwitchCount)
-							.Select((_, i) => new GamelogicEngineSwitch($"ball_switch_{i + 1}", i + 1)
+							.Select((_, i) => new GamelogicEngineSwitch($"ball_switch_{i + 1}")
```
freezy, line 222: Also add a comment so it's clear this is used in the regex.

freezy (commented):
Thanks!

Great to hear that this can be simplified. :)

Will test and look through PinMAME tomorrow.

VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughApi.cs
jsm174: Done. Just curious, if there are multiple troughs on a playfield, would multiple `ball_switch_1`s  cause any issues?

VisualPinball.Unity/VisualPinball.Unity.Editor/Managers/Lamp/LampListViewItemRenderer.cs
jsm174: Done.

VisualPinball.Unity/VisualPinball.Unity/VPT/Trough/TroughComponent.cs
jsm174: Done!

freezy:
Merging this soon. @jsm174 Could you update the MPF repo with these changes as well please?

jsm174:
sure. Will do tonight!

Issue #416, opened 2022-06-06
trough does not get created when table is imported.


shntat:
Trough is missing when table is imported. Have to add manually. trough does eject but fails to eject after ball is drained.

freezy:
We'll have a more complete example table with proper assets and maintained functionality at some point, but right now we're not there yet.

If there is a bug with the trough in general, we need more precise repro steps.

PR #419, opened 2022-06-29
instructions unclear, need additional information and version information Vpinball running-vpe.md

Tom-Neverwinter:
"Now we can begin with some simple game play. Open [Visual Pinball](https://github.com/vpinball/vpinball), create a new "blank" table, and save it somewhere. In Unity, go to *Visual Pinball -> Import VPX* and choose the `.vpx` file you've just created."

I dont understand this instruction, can this be broken down. I am assuming I am importing a .vpx file directly into unity?
"In Unity, go to *Visual Pinball" is this an asset? packages we download in the last section are greyed out. I dont understand the leap from installing to running. 

 I have a variety of Vpinball files is this VP99? VPx? VP1-9? or is it only compatible with Vpinball from that git?

### DIFF
--- a/VisualPinball.Unity/Documentation~/creators-guide/setup/running-vpe.md
+++ b/VisualPinball.Unity/Documentation~/creators-guide/setup/running-vpe.md
@@ -4,7 +4,7 @@ description: How to run VPE
 ---
 # Running VPE
 
-Now we can get begin with some simple game play. Open [Visual Pinball](https://github.com/vpinball/vpinball), create a new "blank" table, and save it somewhere. In Unity, go to *Visual Pinball -> Import VPX* and choose the `.vpx` file you've just created.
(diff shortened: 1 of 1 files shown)


freezy:
If you've never used VPX I agree this is confusing. The final goal of VPE is of course not to create tables in VPX and import them over, but to ship with a reasonable amount of example tables that get you started quickly. So the "getting started" section is more of a temporary thing.

Additionally, since we've started working on a bunch of table builds, the import flow hasn't been tested with the default table for a few months now, and we've had other reports of problems (e.g. #416).

That said, when we talk about *Visual Pinball -> Import VPX*, we mean the new Visual Pinball **menu** that appears when you import the VPE package. The version of VPX isn't important since the file format only slightly changed and we support most of the versions. Just grab the latest release from GitHub.

freezy:
And thanks for the typo fix!

Tom-Neverwinter:
Thank you, I will try and find some time tonight after work and see if I can get this running. 

Saw this a few months back and the pinball nostalgia began, finally found some time this week to try and build it

PR #424, opened 2022-08-12
doc: Add Unity License Secrets Instructions

empty-set:
Unity Secrets must be configured in GitHub before automated builds will complete. Attempting to make this easier for future developers

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,16 +16,16 @@ and let our peers review the code before it gets merged.
 ### Unity License Setup For Automated Testing
 
 All Pull Request must pass an automated run of all the unit tests before merging. These will fail until your fork of the `VisualPinball.Engine` is configured with Unity's license information.
-Most contributos will be using a Personal License and will need to request a key on behalf of GitHub. Professional account users can gather their key from the [Unity Subscriptions Page](https://id.unity.com/en/subscriptions) and skip to create secret setup below
+Most contributors will be using a Personal license and will need to request a key on behalf of GitHub. Professional license users can gather their key from the [Unity Subscriptions Page](https://id.unity.com/en/subscriptions) and skip to step 8.
 
-- Create a new branch. We will use the [Unity - Request Activation File](https://github.com/marketplace/actions/unity-request-activation-file) action to request an activation file
-- Create a file called `.github/workflows/activation.yml` and add the following workflow action defintion.
+1. Create a new branch. We will use the [Unity - Request Activation File](https://github.com/marketplace/actions/unity-request-activation-file) action to request an activation file
+2. Create a file called `.github/workflows/activation.yml` and add the following workflow action defintion.
 ```
 name: Acquire activation file
 on: push
 jobs:
   activation:
-    name: Request manual activation file 🔑
+    name: Request manual activation file 
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


empty-set:
My Build is still failing so I might need to make some modifications to the instructions :) The changes in the PR describe exactly the steps I've taken. These were adapted from this link. https://game.ci/docs/github/activation#personal-license I'm using a personal license, and I did receive a proper ulf.

jsm174:
hmm. I think maybe we should just include the activation workflow in the repo, so we don't have to make the instructions so complicated. I'll look in it.

empty-set:
> hmm. I think maybe we should just include the activation workflow in the repo, so we don't have to make the instructions so complicated. I'll look in it.

I would be ok with this, the workflow will need to be modified to not include the on: push since it will only need to run once.

The original instructions suggest using on: workflow_dispatch: {} and manually running the action. I was unable to manually run actions in GitHub,  my UI in GitHub actions does not match the screenshots. 

Also my build is still failing so my instructions must also be missing something?

CONTRIBUTING.md
```diff
@@ -13,6 +13,49 @@ and let our peers review the code before it gets merged.
 - We try to prefix our commit messages with a word that quickly tells the reader where the change happened. Examples are `editor`, `doc`, `<component-name>`, etc.
 - Changes that touch the core project should be unit-tested. 
 
+### Unity License Setup For Automated Testing
+
+All Pull Request must pass an automated run of all the unit tests before merging. These will fail until your fork of the `VisualPinball.Engine` is configured with Unity's license information.
```
freezy, line 18: I would formulate is less restrictive. We'll still merge PRs when it's just doc or something that we test locally. Maybe:

```suggestion
It's preferred to make the automated tests run when creating your PR. Since Unity needs a license key (which can be obtained with a free account), you'll need to configure your fork to use the correct secrets.
```

CONTRIBUTING.md
```diff
@@ -13,6 +13,49 @@ and let our peers review the code before it gets merged.
...
+6. Visit (license.unity3d.com) and upload the `Unity_v20XX.X.XXXX.alf` file.
+7. You should now receive your license file (Unity_v20XX.x.ulf) as a download. It's ok if the numbers don't match your Unity version exactly.
+8. Open `Github` > `<Your repository>` > `Settings` > `Secrets`
+- Create the following secrets
+  - `UNITY_EMAIL` - (Add the email address that you use to login to Unity)
+  - `UNITY_PASSWORD` - (Add the password that you use to login to Unity)
+- Personal License
+  - `UNITY_LICENSE` - (Copy the contents of your ulf license file into here)
+- Professional License
+  - `UNITY_SERIAL` - (Add you serial key it should look like XX-XXXX-XXXX-XXXX-XXXX-XXXX)
```
freezy, line 54: After talking with @jsm174, the preferred method would be to actually include `activation.yml` in the repo so the user must only set the correct secrets instead of additionally copying everything over.

freezy (changes requested):
Thanks!

@jsm174 will add the necessary so steps 1-3 become unnecessary.

empty-set (commented):
Looks good to me, when I tried the workflow_dispatch previously, I was not able to run the workflow manually. I'll give it a retry

PR #429, opened 2022-08-20
Feature/flipper tricks 2

Cupiii:
This adds FlipperTricks by RothbauerW.
I implemented this exactly like RothBauerW did in his script (same variable names).
There is no microbounce to the ball yet, if releases an ball is in cradle - this will come together with dropcatch.

To test, please also switch branch at the table.

FlipperTricks is off by default (on new Flippers).

This PR replaces the old PR: https://github.com/freezy/VisualPinball.Engine/pull/404


Cupiii:
Oh, @freezy: I had some .meta-Files deleted - not sure this is right, and did not know how to revert it back. Also changes Gitignore... this may be wrong. If you could please have a look?

freezy:
Awesome, thanks! Will have a look.

jsm174:
We definitely need to revert any deletions to meta files in the plugins folder, otherwise the ci will be busted. Also there have been a few times where we legitimately needed to update those metas (all the cross platform stuff), so not sure if adding them to .gitignore. I guess we could just force commit the files?

Cupiii:
That's what I also thought. I can try to revert them, but I don't really know how to force push. Or can you/freezy do that/both?

Cupiii:
I tried to get the giles back, but I faild within my git gui. So if anybody vould resolve the issue, that would be great.

freezy:
Don't worry about it, I'll do it in a minute.

For the future, in order to avoid committing them when Unity removes them, check [this](https://github.com/freezy/VisualPinball.Engine/wiki/Development-Setup#meta-files).

Cupiii:
Ok, Thanks, so I have to put these 18 files to the "ignore" via "assume-unchanged"? Will do, thx.

freezy:
Yes, because you can't `.gitignore` deleted files unfortunately.

freezy:
Should be fixed now. I've also rebased on master. `git fetch && git reset --hard origin/feature/Flipper_Tricks_2` will get you back sync.

Cupiii:
Ah, great, thanks. 

I may have broken something with the bounding boxes. Let me check tomorrow.

Cupiii:
> I may have broken something with the bounding boxes. Let me check tomorrow.

Nothing broken, I think. It was broken in the middle of the implementation and i put it on my todo. I got a null-referce error. But everything seems to be OK, just tested. Bounding boxes are bigger because of the overshoot - that's intended.  

Also reviewed everything again. I can't see anything strange.

PR #430, opened 2022-08-20
Additionally map string-parsed IDs when mapping coils

freezy:
There was a regression introduced when we cleared internal (`int`) IDs in the GLE in favor of purely string-based IDs.

The coil player currently blindly looks up the ID it gets from the GLE, which in the case of PinMAME, is a stringified integer. However, we often prefix coil IDs with "0" to get proper sorting, and it's also how they are named in the manual. So a "04" coil mapping won't match a "4" event from the GLE.

This patch additionally maps coils to their stringified integer, if it's not the same as the original name.

There will be another problem, the other way around, for switches. I will update this PR with a fix for that too, when I can reproduce.

I'll also do the same for lamps. @jsm174 let me know your thoughts already.


freezy:
Also need to check whether that doesn't screw up the coil statuses, because `CoilStatuses` is not distinct anymore.

jsm174:
> However, we often prefix coil IDs with "0" to get proper sorting

The treeview control used for switch and coil managers was update to support sorting by number here:

https://github.com/freezy/VisualPinball.Engine/commit/58c77cf90fc1f3ddbc73d959cb6ec317516ad68e

freezy:
Ah, cool. Still want to support people prefixing names with "0" though.

PR #435, opened 2022-08-26
Implement Score Motor component

### LINKED ISSUE #421: Properly simulate score reels
We already have a score reel component, but it's very simplistic. In fact, a lot more is going on in a score reel, and the goal of this issue is to specify it and find a solution to implement it in VPE, as well as expose it through Visual Scripting.

## Single Point vs Multi Point

We call *single point scoring* when a reel only advances by one position. It doesn't matter which reel, so a score changing from 0 to 100 or from 0 to 1000 is both called single point scoring. If multiple reels change one position at the same time (e.g. from 0900 to 1000), we also call single point scoring.

## Direct Switch vs Motor-Driven Switches

Moving a reel to the next position is done through a coil. For single scoring elements on the playfield, a switch triggers the appropriate coil directly. For multi scoring, a motor activates the switches.

Please check [this excellent website](https://www.funwithpinball.com/learn/animated-score-motor-circuits) to understand how score motors work.

## Blocking

While the motor is running, it blocks additional points during that time. For example, if the score goes from 0 to 500, and while the motor is moving the hundreds from 0 to 5, another 100 is scored, that 100 is discarded. If however a single point is scored on another reel (for example, a 10), it might be accepted (but a 50 wouldn't be accepted since the motor is busy moving the hundreds from 0 to 500).

## Resetting

Resetting starts the score motor, pulses the reels until 0 is reached on all, then turns off the score motor. For example, a score of 2041 would cause the following sequence: 3052, 4063, 5074, 6085, 7096, 8007, 9008, 9, 0.

No more than two score banks can be reset a time. On a 4 player game, one and two will reset, followed by three and four. 

## Side effects

The motor status is globally of interest, since some tables like Volley turn off the bumper and insert lights while the motor is running.

# VPE Approach

We want to provide a simple way for creators to do score-reel scoring, but we also want more advanced creators give the possibility to accurately simulate a score reel with a score motor.

The simple way is just adding a number to the score and sending the score to the display. The display (in our case, the score reels, but the same would apply to a segment display as well) would thus receive an integer value and doe what it needs to do autonomously.

For the accurate way, we'll provide a *Score Motor Component*. This component comes with a (pulse) coil input to turn the motor on, and a configurable number of switch outputs. Inside the component, one can configure how the score motor is set up, i.e. how many cams it features, and the pulse patterns for each switch. (It might be easier to assign a pattern to each switch, instead of cams and the position of the switches on each cam.)

Then, our `ScoreReelComponent` would become a `ICoilDeviceComponent` that takes in a coil input which moves the reel to the next position (it might be better to make `ScoreReelDisplayComponent` a `ICoilDeviceComponent`, since that would allow easier tracking of the score).

So for advanced mode, one would add the new `ScoreMotorComponent`, link its switch outputs to the coil inputs of the `ScoreREelDisplayComponent` through the wire manager, and use the reel- and motor switches in the GLE to update the score, instead of an integer.

jsm174:
This PR closes #421.

It adds a new `ScoreMotorComponent` that can be configured to correctly one or many `ScoreReelDisplayComponent`.  

`ScoreReelDisplayComponent` has been updated to look for a `ScoreMotorComponent`. If found, the display will keep track of its score.

The Score Motor can be configured in the inspector:

Screen_Shot_2022-08-26_at_5 29 26_PM

The default settings are suitable for a Gottlieb EM, ie:  6 steps in 120 degrees over 769ms.

The component forces a working configuration, ie, you can not configure less than 5 steps.

The `ScoreMotorComponent` also exposes two switches `Motor Running Switch` and `Motor Step Switch` that can be configured in the `Switch Manager`:

Screen_Shot_2022-08-26_at_5 28 51_PM

There will be a follow up PR that adds new Visual Scripting Units to tie this all together.

Thanks again to @freezy, @Scottacus64, @bord, and @Cupiii for the several hours of discussion to get this figured out.


Scottacus64:
This looks awesome!  Thanks for all of the hours you put into this!

jsm174:
After discussions on Discord, @freezy suggested using the existing `Update Display` unit instead of an additional `Add Points` unit.  

I've refactored the code accordingly, and it really makes everything a lot more clean. The `ScoreReelDisplayComponent` now keeps track of its score internally.

I've also added selecting the `ScoreMotorComponent` directly on the `ScoreReelDisplayComponent`. This would allow us to have multiple score motors with different configuration for testing:

Screen Shot 2022-08-28 at 2 11 08 PM

If a `ScoreMotorComponent` is not attached to the `ScoreReelDisplayComponent`, logic works exactly like it did prior to this PR.

jsm174:
The last remaining issue is to revisit the score reel animation. It can often times get out of sync with the reels internal score.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+		[Tooltip("Disable to allow single point scores while score motor running.")]
+		public bool BlockScoring = true;
+
+		public List<ScoreMotorActions> ScoreMotorActionsList = new List<ScoreMotorActions>()
+		{
+			new ScoreMotorActions(),
+			new ScoreMotorActions(),
+			new ScoreMotorActions(),
+			new ScoreMotorActions(),
+			new ScoreMotorActions()
```
freezy, line 103: Shouldn't there be 6 of them, given `Steps` is 6 per default?

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+		};
+
+		public const string MotorRunningSwitchItem = "motor_running_switch";
+		public const string MotorStepSwitchItem = "motor_step_switch";
+
+		public event EventHandler OnUpdate;
```
freezy, line 109: No args here?

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+		public event EventHandler<ScoreMotorResetEventArgs> OnReset;
+		public event EventHandler<ScoreMotorAddPointsEventArgs> OnAddPoints;
```
freezy, line 111: Still needed? Or can we do `OnSetScore` as well?

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+		IEnumerable<GamelogicEngineSwitch> IDeviceComponent<GamelogicEngineSwitch>.AvailableDeviceItems => AvailableSwitches;
+
+		#region Runtime
+
+		private void Awake()
+		{
+			GetComponentInParent<Player>().RegisterScoreMotorComponent(this);
+		}
+
+		private void Update()
```
freezy, line 135: Where is this called?

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
```diff
@@ -0,0 +1,277 @@
...
+		public event EventHandler Init;
+
+		IApiSwitch IApiSwitchDevice.Switch(string deviceItem) => Switch(deviceItem);
+
+		public DeviceCoil ResetCoil;
+		public DeviceSwitch MotorRunningSwitch;
+		public DeviceSwitch MotorStepSwitch;
+
+		private int _degreesPerStep;
+		private float _degreesPerSecond;
```
freezy, line 44: Can both be read-only.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
```diff
@@ -0,0 +1,277 @@
...
+		private readonly ScoreMotorComponent _scoreMotorComponent;
+		private readonly Player _player;
+
+		public event EventHandler Init;
+
+		IApiSwitch IApiSwitchDevice.Switch(string deviceItem) => Switch(deviceItem);
+
+		public DeviceCoil ResetCoil;
+		public DeviceSwitch MotorRunningSwitch;
+		public DeviceSwitch MotorStepSwitch;
```
freezy, line 41: Can be private

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
```diff
@@ -0,0 +1,277 @@
...
+		private static readonly Logger Logger = LogManager.GetCurrentClassLogger();
+
+		private readonly ScoreMotorComponent _scoreMotorComponent;
+		private readonly Player _player;
+
+		public event EventHandler Init;
+
+		IApiSwitch IApiSwitchDevice.Switch(string deviceItem) => Switch(deviceItem);
+
+		public DeviceCoil ResetCoil;
```
freezy, line 39: Unused?

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
freezy: Okay I got it, this is for relaying Unity's `Update` to the API.

However I think so far, when frame-bound logic was needed during runtime, that logic would be in the Component, not the API class. The API classes should be detached from Unity as much as possible.

freezy: Hmm but in this case, we really just need the tick. Usually we move stuff around in the scene, but this should be fine. Maybe put a comment, I took this for updating the score or something.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
```diff
@@ -0,0 +1,277 @@
...
+				((e.Points % 1000000000 == 0) ? e.Points / 1000000000 :
+				 (e.Points % 100000000 == 0) ? e.Points / 100000000 :
+				 (e.Points % 10000000 == 0) ? e.Points / 10000000 :
+				 (e.Points % 1000000 == 0) ? e.Points / 1000000 :
+				 (e.Points % 100000 == 0) ? e.Points / 100000 :
+				 (e.Points % 10000 == 0) ? e.Points / 10000 :
+				 (e.Points % 1000 == 0) ? e.Points / 1000 :
+				 (e.Points % 100 == 0) ? e.Points / 100 :
+				 (e.Points % 10 == 0) ? e.Points / 10 :
+				 e.Points);
```
freezy, line 131: I don't understand this block. You're basically removing zeros from the score to add. But that score can be something like 1500, no? In this case you'll get 1 and drop the 500?

freezy (commented):
Thanks, this is so cool!

Added a few comments. One thing I still don't understand is why can't you completely replace the *add* points methods by *set* point methods? Then you can also unify it with *reset*. I still don't get why reset is so different from "set points to 0".

Awesome stuff though. Will check out the UVS part tomorrow.

VisualPinball.Unity/VisualPinball.Unity/Display/DisplayComponent.cs
```diff
@@ -25,6 +26,8 @@ namespace VisualPinball.Unity
 {
 	public abstract class DisplayComponent : MonoBehaviour
 	{
+		public DisplayPlayer _displayPlayer { get; set; }
```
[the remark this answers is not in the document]
jsm174: Done

VisualPinball.Unity/VisualPinball.Unity/Display/ScoreReelDisplayComponent.cs
```diff
@@ -37,20 +45,59 @@ public class ScoreReelDisplayComponent : DisplayComponent
...
+				UpdateFrame();
+			}
 		}
 
 		public override void UpdateFrame(DisplayFrameFormat format, byte[] data)
 		{
-			var score = (int)BitConverter.ToSingle(data);
-			var digits = DigitArr(score);
+			if (ScoreMotorComponent) {
+				var points = BitConverter.ToSingle(data);
```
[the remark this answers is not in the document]
jsm174: Done

VisualPinball.Unity/VisualPinball.Unity/Game/Engine/IGamelogicEngine.cs
```diff
@@ -191,6 +194,16 @@ public enum DisplayFrameFormat
 		Numeric       // gets a byte-array converted float
 	}
 
+	public class DisplayClearData
```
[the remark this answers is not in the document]
jsm174: Done. I just pass this as a string id.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+	{
+		Wait = 0,
+		Increase = 1
+	}
+
+	[Serializable]
+	public class ScoreMotorActions
+	{
+		public List<ScoreMotorAction> Actions = new List<ScoreMotorAction>();
+	}
```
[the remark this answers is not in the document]
jsm174: Done. Moved to the bottom.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,152 @@
...
+
+	[AddComponentMenu("Visual Pinball/Game Item/Score Motor")]
+	public class ScoreMotorComponent : MonoBehaviour, ISwitchDeviceComponent
+	{
+		public const int MaxIncrease = 5;
```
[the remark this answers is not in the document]
jsm174: During dev, I was discussing with @Scottacus64. The increases in a EM are never greater than a multiple of 5, ex: 1, 2, 3, 4, 5, 10, 20, 30, 40, 50, 100, 200, 300, 400, 500, 1000, 2000, 3000, 4000, 5000, etc. 
 
So you can't score 1202 points.  

I didn't want to have 5 all throughout code, so I figured I'd make it a constant.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
jsm174: I couldn't figure out how to do multi-dimensional arrays with `SerializedProperty`, so I had to make a list of a `ScoreMotorActions` classes that had a list of `ScoreMotorAction` enums.

To make this a little clearer, I renamed the class to `ScoreMotorTiming`

So we will always have 5 timings, with each one having "step" amount of actions.

It would be nice if I could have done this with something like:

`public List<ScoreMotorTiming> ScoreMotorTimingList = Enumerable.Repeat(new ScoreMotorTiming(), ScoreMotorComponent.MaxIncrease).ToList();`

But then the timing list is reset anytime the inspector is open.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
jsm174: Sorry, I was a little confused about API vs component.  I will move everything into the component.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
jsm174: Done

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
jsm174: Done

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
jsm174: yup. something I forgot to remove. Done.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
jsm174: This goes back to the `MaxIncrease` question / comment.  The final value calculates as 1, 2, 3, 4, or 5. 

On an EM with a score motor, you could never score 1500. @Scottacus64 can confirm this.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
jsm174: Okay, I moved everything out of API into the component. That got rid of those extra `ScoreMotor*EventArgs` classes.

Done

jsm174:
> Added a few comments. One thing I still don't understand is why can't you completely replace the _add_ points methods by _set_ point methods? 

The reason why I'm using add points is because of the blocking. 

For example, we have 1000 points and we tell the score display to update 500 points:

- score motor starts running
- first increase of 100, the internal score is 1100
- incoming points request for 10 points and blocking disabled, we add 10 points, the internal score is 1110
- second increase of 100, the internal score is 1210
- incoming points request for 500 points, and we throw it out 
- incoming points request for 10 points and blocking disabled, we add 10 points, the internal score is 1220
- third increase of 100, the internal score is 1320
- fourth increase of 100, the internal score is 1420
- fifth increase of 100, the internal score is 1520
- score motor stops running

> Then you can also unify it with _reset_. I still don't get why reset is any different from "set points to 0".

Since we had the Update Display visual scripting unit, and each display offers a clear method, I thought a Clear Display unit would be a good addition. 

Reset is doing a little more than just setting the reels to zero.  It follows all the same timing rules, and has logic to keep the motor running for an additional turn if needed.

Say the score is 2230, one turn is going to advance the digits 5 times:

3340, 4450, 5560, 6670, 7780

We still need to respect that timing (wait, increase, increase, increase, increase, increase). 

Since we are not at zero, the score motor continues for another revolution 

8890, 9900, 0, 0, 0 

(wait, increase, increase, increase, increase, increase).

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
jsm174: No longer needed as moved from Api to inside component. Done.

VisualPinball.Unity/VisualPinball.Unity/Game/Engine/IGamelogicEngine.cs
```diff
@@ -42,7 +42,10 @@ public interface IGamelogicEngine : IGamelogicBridge
 		/// most GLEs only know about their displays when they start the game.
 		/// </remarks>
 		event EventHandler<RequestedDisplays> OnDisplaysRequested;
-		event EventHandler<DisplayFrameData> OnDisplayFrame;
+		event EventHandler<DisplayClearData> OnDisplayClear;
+		event EventHandler<DisplayFrameData> OnDisplayUpdateFrame;
+
+		void DisplayUpdateEvent(DisplayFrameData displayFrameData);
```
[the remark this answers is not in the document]
jsm174: I added a doc block. I'd still like the names be a little cleaner. Will work on that tonight. 

In the gles, we have 

```
		public event EventHandler<LampEventArgs> OnLampChanged;
		public event EventHandler<LampsEventArgs> OnLampsChanged;
		public event EventHandler<CoilEventArgs> OnCoilChanged;
		public event EventHandler<SwitchEventArgs2> OnSwitchChanged;
```

I'd like to have something like

```
		public event EventHandler<DisplayEventArgs> OnDisplayChanged;
```

I'm not seeing the connection with the DisplayPlayer, but I will keep looking.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
Scottacus64: Yes only multiples of 1, 10, 100, 1000 etc.  There would be no way for scoring three 500's, or five 300's.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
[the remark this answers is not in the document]
Scottacus64: @jsm174 you've got a good handle on how a score motor works!

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
freezy: Got it. Although in this case, this block could be solved more elegantly, no? E.g.

```cs
var increase = Math.Floor(e.Points / Math.Pow(10, Math.Floor(Math.Log10(e.Points))));
```

freezy:
Good point, I didn't think about the concurrency issue. Now it makes sense!

I'll go through it again tonight and merge if all is good.

Thanks for bearing with me :)

jsm174:
> Good point, I didn't think about the concurrency issue. Now it makes sense!
> 
> I'll go through it again tonight and merge if all is good.
> 
> Thanks for bearing with me :)

I went through and cleaned up the `DisplayUpdateEvent` to just `DisplayEvent`.  Makes it more like `SwitchEvent` `CoilEvent` etc.   

I also removed `DisplayPlayer` from `Displays`. They now use an `OnDisplayChanged` event.

jsm174:
> Oh, one more thing.
> 
> I suppose that currently, in UVS, it's entirely possible for a user to put scores like 1500. How do we deal with that? Error, warning, or fall back to non-score-motor driven scoring for those cases?

The score motor figures out the increase doesn't make sense and then logs the error, and the points are thrown away.

```
			if (increase > ScoreMotorComponent.MaxIncrease) {
				Logger.Error($"too many increases (ignoring points), id={id}, points={points}, increase={increase}");
				return;
			}
```

VisualPinball.Unity/VisualPinball.Unity/Game/Engine/IGamelogicEngine.cs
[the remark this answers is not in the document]
jsm174: This is cleaned up, so it would be easier to re-review the file as a whole.  I added an `SetDisplay()` in the gle which gets called with an `OnDisplayChanged`. This makes it feel more like the rest of the code.

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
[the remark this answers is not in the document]
jsm174: thanks @Scottacus64. Only because of you guys! 😄

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorApi.cs
jsm174: I'm not good with all that fancy math 😄. 

Updated: 

`
var increase = (int)System.Math.Floor(points / System.Math.Pow(10, System.Math.Floor(System.Math.Log10(points))));
`

VisualPinball.Unity/VisualPinball.Unity/VPT/Mech/ScoreMotorComponent.cs
```diff
@@ -0,0 +1,293 @@
...
+
+						return;
+					}
+				}
+
+				StopMotor();
+			}
+		}
+
+		public void Reset(string id, float score, ScoreMotorResetCallback callback)
```
freezy, line 142: I'll rename this to `ResetScore`, since [Reset is already taken](https://docs.unity3d.com/ScriptReference/MonoBehaviour.Reset.html).

freezy (approved):
Okay for me now.

In the future, to do timing-related stuff, [coroutines](https://docs.unity3d.com/Manual/Coroutines.html) might be an easier approach.

Can be merged once the doc is done.

freezy:
Oh, and don't forget the UVS doc for the new nodes!

freezy:
Also, PinMAME needs to be updated with the new API (and preferably MPF as well)

PR #436, opened 2022-08-27
Feature/live catch

Cupiii:
Adds live catch to flippers.

There are some differences to nFozzy:
1. Max and min distance is not calculated on x axis, but on tangent to the flipperface from the flippers base. (so that live catches could probably also be done with vertical flippers) 
2. The minimal ball speed for a live catch is a variable and not a constant in the code
3. Not only the time for live catches (from flipped flipper to collision) can be set, but also the minimal time for a "perfect" live catch. (was always 0.5*liveCatchTime)
4. perfect live catches were always zero velocity at nFozzy's, now you can set a multiplicator (default = 0.1), so even a perfect catch has some little bounce - this adds some realism, but can be turned off, when setting is set to 0.0.
5. imperfect live catches added some constant velocities at nFozzy's (depending on the time). Now a multiplicator can be set. maximum inaccurate bounce multiplicator to original speed is applied at full time. Linear regessed to perfect bounce multiplicator, when getting to perfect time. (like at nFozzy).
6. Rubber and the collision to it is still calculated for imperfect (or perfect bounces). 
7. Z-Rotation is not nulled. A rotating ball can rotate from the flipper, as seen in live-catch videos.
8. Every change to the speed of the ball is calculated with the normal direction of the flipper face in mind. So there is no live catch for balls that come from wrong diagonal (for left flipper: fast balls coming from top right had very strange bounces at nFozzy's).  

The default vales differ from nFozzy's script, because of 1 and 5.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperCollider.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperCollider.cs
@@ -722,12 +722,17 @@ namespace VisualPinball.Unity
 		#region LiveCatch
 		public void LiveCatch(ref BallData ball, ref CollisionEventData collEvent, ref FlipperTricksData tricks, in FlipperStaticData matData, uint msec )
 		{
+			var normalSpeed = math.dot(collEvent.HitNormal, ball.Velocity) * -1f;
 
-			if (ball.Velocity.y > 6)
-			{
-				Logger.Info("LiveCatchTest - Ball with y-speed {0}, at CollisionTime: {1}, livecatchTime is {2}, difference is {3} msecs", ball.Velocity.y, msec, tricks.FlipperAngleEndTime*1000, tricks.FlipperAngleEndTime*1000-msec);
-				ball.Velocity.y = 0.6f;
-			}
+			// only test for LiveCatch if Ballspeed is greater as set Minimal Speed (default = 6)
+			if (normalSpeed >= tricks.LiveCatchMinimalBallSpeed)
+
+				if (ball.Velocity.y > 6)
+				{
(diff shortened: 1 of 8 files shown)


Cupiii:
Argh. Forgot to check if the checkbox is checked... Will add that tomorrow.

freezy:
Amazing! This is on my to-review list today :)

Cupiii:
> Amazing! This is on my to-review list today :)

You can test it quite well, if you set the perfect time to 0.5 secs and full time to 1s. It is "just a tad" easier then :-)

VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperCollider.cs
```diff
@@ -717,6 +718,56 @@ private void GetRelativeVelocity(in float3 normal, in BallData ball, in FlipperM
 
 		#endregion
 
+		//public static readonly Logger Logger = LogManager.GetCurrentClassLogger();
+		#region LiveCatch
+		public void LiveCatch(ref BallData ball, ref CollisionEventData collEvent, ref FlipperTricksData tricks, in FlipperStaticData matData, uint msec ) {
+			if (!tricks.UseFlipperLiveCatch)
+				return;
+			var normalSpeed = math.dot(collEvent.HitNormal, ball.Velocity) * -1f;
```
freezy, line 732: We can move this calculation down a bit after the early exits, right?

freezy (approved):
Fantastic stuff! I've pushed a few minor changes, such as variable naming, added units to the inspector, and made one method static.

Cupiii:
"Bump on release" has to be greyed out or commented out, since not yet implemented (comes with flipper-bounce in some weeks).  Will add a commit for this next day(s). 
According to discussion on discord, will eventually refactor this into presets.

Issue #449, opened 2023-01-14
Handle Mechanical Sounds


freezy:
This is a proposition how to handle mechanical sounds, described in #226. There are a few additional constraints that aren't described in that issue that made me go back and forth on a few different approaches. But here is what I think could work.

## Sound Definition

As described in #226, a sound can contain multiple clips, random pitch and so on. So the first step is to define such a sound asset, typically as a [`ScriptableObject`](https://docs.unity3d.com/Manual/class-ScriptableObject.html).

In the inspector, it could look something like this:

<img width="306" src="https://user-images.githubusercontent.com/70426/212437866-233f22d3-9a5a-4c54-bc6d-65f97aa2b1e3.png">

These ScriptableObjects are files that reference audio clips, and they sit on the file system (as opposed to the scene), which makes them also usable as assets in the asset library.

## Mapping

One of the main concerns is that we don't want authors to have to manually assign let's say a collision sound to every single element that collides with the ball. Given we have an asset library, the sound should already be part of a prefab. Of course, it should be possible to assign different sounds, but assets should be ready-to-go where possible.

Which means that the sound configuration should be GameObject-specific. Let's start with a simple case:

<img width="306" src="https://user-images.githubusercontent.com/70426/212438327-b12fa53c-2b8f-4a76-90e0-e08d09e8e536.png">

So this is the inspector of a Flipper object, where we see a new *Mechanical Sounds* component. In here, we have a list of sounds that to be emitted on various actions.
- The first sound is the flipper coil when enabled. The **Trigger** property looks up which trigger events this GameObject supports, and presents them in a drop-down.
- The **Sound** property is a reference to the sound asset we defined above.
- The **Volume** property defines how to set the volume. *Fixed* set to `1` means same volume as configured by the sound asset. This includes the asset's own volume correction and randomization, if set. The slider just acts a multiplicator, allowing fine-tuning on a per-object basis.
- **Action** defines, well, what to do (start, loop, stop).
- Finally, the **Fade** property sets a time for how long the sound should be faded in or out.

In this example, there are three sounds:

1. Play the "Flipper Coil On" sound once when the coil of the flipper turns on.
2. Start looping the "Flipper Coil Hum" sound when the coil of the flipper turns on.
3. Stop looping the "Flipper Coil Hum" sound when the coil of the flipper turns off.

[...]

PR #453, opened 2023-02-13
Add mechanical sound support.

Mr-Jay-Gatsby:
This is the initial commit for the mechanical sounds as discussed here:
https://github.com/freezy/VisualPinball.Engine/issues/449.

Main items added are:
Sound Asset(scriptable object)
Sound Inspector
Mechanical Sounds Data
Mechanical Sounds Component
Mechanical Sounds Inspector
ISoundEmitter


Mr-Jay-Gatsby:
When working in the specific table project, I did add an AudioMixer asset called SoundMixer to try and alter the speed of sounds. I exposed the pitch shifter parameter and called the mixer in the sound inspector script but the setfloat method wasnt working (getfloat did work). I dont know another way to change the speed unless there are some external plugins to use.

VisualPinball.Unity/VisualPinball.Unity.Editor/VPT/Sounds/SoundInspector.cs
```diff
@@ -0,0 +1,357 @@
...
+		}
+		
+		private void Update()
+		{
+			
+			if (_loop)
+			{
+				//set up new array of gameobjects based on most current number of audioclips, if new loop is initiated
+				if (_soundObjects == null || _soundObjectsLength == 0) 
+				{ _soundObjects = GetSoundObjects();}
```
freezy, line 312: What's the reason for this? Is it impossible to schedule it on the same audio source?

I'll need to dig more into Unity's audio system, but IMO you should have an audio source on each GameObject that emits audio, and play the also the editor sound through that. (Adding the audio source automatically can be easily achieved with `[RequireComponent(typeof(AudioSource))]` on `MechSoundsComponent`.)

VisualPinball.Unity/VisualPinball.Unity/VPT/Flipper/FlipperApi.cs
```diff
@@ -17,6 +17,7 @@
 // ReSharper disable EventNeverSubscribedTo.Global
 #pragma warning disable 67
 
+using Codice.CM.SEIDInfo;
```
freezy, line 20: Some IDE garbage, I assume? :)

VisualPinball.Unity/VisualPinball.Unity/VPT/Sounds/MechSounds/MechSoundsComponent.cs
```diff
@@ -0,0 +1,209 @@
...
+			{
+				volEmitter = new VolumeEmitter();
+				volEmitter.Id = Ids[i];
+				volEmitter.Name = Names[i];
+				volEmitters[i] = volEmitter;
+			}
+
+			return volEmitters;
+		}
+		#endregion
```
freezy, line 117: All this should be in the individual item components and is probably not necessary anymore.

VisualPinball.Unity/VisualPinball.Unity/VPT/Sounds/MechSounds/MechSoundsComponent.cs
```diff
@@ -0,0 +1,209 @@
...
+
+
+		public override IEnumerable<MonoBehaviour> SetReferencedData(MechSoundsData data, Table table, IMaterialProvider materialProvider, ITextureProvider textureProvider, Dictionary<string, IMainComponent> components)
+		{
+		
+			return Array.Empty<MonoBehaviour>();
+		}
+
+
+		#endregion
```
freezy, line 176: Most of this should be moved to individual items components as well.

VisualPinball.Unity/VisualPinball.Unity/VPT/Sounds/MechSounds/MechSoundsComponent.cs
```diff
@@ -0,0 +1,209 @@
...
+			
+		}
+
+		#endregion
+
+		#region Editor Tools
+
+
+
+		#endregion
```
freezy, line 206: Idem :)

freezy (commented):
Great stuff, thanks a lot!

You'll see in my comments that most of the code in `MechSoundsComponent` should be moved to the individual components. I'd suggest to just do it for the flipper, then I'll help doing the rest of the items.

There are some minor code style incoherences that I'll comment on as this PR matures.

All in all, awesome work. Looking forward to getting this in. :)

freezy:
I've pushed a few commits. Changes:

- Use a static audio source for the editor instead of temp creating one on play.
- Let Unity handle looping, since `AudioSource` supports it
- Dropped "Speed" for now, since it needs a mixer hooked to an effect
- Ditched the volume emitter, since it's actually not needed (volume is defined at runtime when the event occurs).
- Move classes into separate files and all to a common folder
- Renamed variables and members to existing conventions
- Simplified `MechSoundDrawer` and `MechSoundInspector`

Still TODO:

- [ ] Propagate and velocity of the collision
- [ ] Do the other game items
- [ ] Add proper icons

linojon:
Hello. I've resumed work on this feature. Work in progress resides in my fork branch https://github.com/linojon/VisualPinball.Engine/tree/feature/sounds . I am building it into the Volley table  (https://github.com/jsm174/vpe-volley/tree/sounds ). using that as a baseline project and trying get our sounds like the VPX version (https://vpuniverse.com/files/file/8185-volley-gottlieb-1976/ ).

Changes so far:
- Implemented mechanical sound Stop action (e.g. for looping sounds)
- Fixed support for multiple concurrent sounds from same game object (requires multiple Audio Sources on the object).
- Optionally give an Audio Mixer, else it searches for one in the nearest parent up the scene hierarchy.
- When adding a sound clip Unity to the mech sounds list, Unity ignores default values so i added an Add button that initializes Volume to 1
- Verified sound emitter support in Flippers, Bumpers, Kickers, Plunger
- Added sound emitter support to Targets (Hit and Drop), Trigger Switches

To test and verify, i am using the vpe-volley project,
- Added the Mechanical Sounds component to bumpers (3X), flippers (2X), targets (yellox 5X, blue 5X, green 5X), kickers (drain, ballRelease), plunger, triggers (9X)
- Added corresponding Sound Assets
- Note, i'm only adding the sounds actually used in VPX script which is a subset of the sounds actually in that table.

Sounds todo:
- Drop target bank reset
- Chimes
- Mechanical score wheel
- Bonus score
- Collect points
- New game 
- High score initials

Plus object collision and ball rolling sounds. I'll make a separate github issue  https://github.com/freezy/VisualPinball.Engine/issues/463 


- 


-

linojon:
i've committed my current work in progress. I have questions, could use some help as i'm still just getting familiar with how VPE works.

* as mentioned, we now have sounds for flippers, bumpers, kickers, plunger, hit targets, drop targets, switches.

* Drop target reset: there's 2 ways to do this. For each target, play a reset sound. I have this implemented. See DropTargetComponent.cs and DropTargetApi.cs. So that's 15 separate simultaneous sounds (i dont know if that matters but just saying). 

Or, we could play one sound for the whole target bank reset (this is what the VPX game does). I tried this and it works but the code is messy, as I needed to add ItemApi to DropTargetBankApi.cs, which (in order to reference local variable MainComponent) requires i derive DropTargetBank from something like MainRenderableComponent, which in turn required i add a bunch of unused methods. See DropTargetBankComponent.cs. If this is a bad idea, i'll clean up the code.

* Score Motor / Reel sounds: as with DropTargetBank, does not recognized MainCompponent I'm guessing the sound would be added to ScoreMotorComponent. Currently this is commented out, see ScoreMotorComponet.cs, ScoreMotorApi.cs. Note that the motor  component is on the Volley game object, which seems pretty high up in the hierarchy to add mechanical sounds.

* Chimes: VPX script triggers chimes when the score is incremented (see sub AddScore1, sub AddScore2). Whats the best place to add chimes in VPE? I expect we add it to the specific score reel (10's, 100's, etc)? Currently, the ScoreReelComponent is a MonoBehaviour and doesnt have any interfaces. 

* Start Game: I tried trace what happens when you press "1" to start a new game, so i can play a new game sound (score motor reels and targets resetting). Maybe the reset sounds belongs in SwitchPlayer?  or by the respective objects?

* Also could use advise how to add sounds for ball rolling and collisions. Try to use ISoundEmitter for this too or something different?

freezy:
Awesome! Could you open a PR from your branch to `feature/sounds`? This branch is the rebased version of this PR, that would allow to properly review and at the same time keep @Mr-Jay-Gatsby's work. I'll then close this one here and we can continue discussing there.

About your questions:

- Drop targets: TBH I'm not sure what's best. I could see both approaches work. They probably won't work if there's only one sample for each reset sound, but if there are multiple, it might work. I haven't looked at your code yet.
- Score Motor: Yes, we'll need a score motor `MainComponent` for this. I think this might be a new use case where we have a non-renderable object that still needs to be positioned in the scene so the spatial sound works. So, there's probably some refactoring needed.
- Chimes: I think chimes can be used for all kinds of things, not just scores. So that'd be something to be linked through the (visual) table script. So we'd need a way to create sound emitters on the playfield that we can reference in visual scripting. Probably the same as the last point.
- Start Game: This should come from the score motor and target components, right? Independently whether a game started or in-game.
- Ball rolling is out of scope for this one, since it's not linked to any components (well, the ball, but it's more complex than the rest). Collision sounds need the correct data from the physics engine. I'll look into that.

Apologies for not being 100% in the context yet. I'll need to re-read the code that has been written so far to be of more use. Many thanks already for your contributions!

freezy:
Closed by #506.

PR #466, opened 2024-01-08
More mechanical sound support

linojon:
Work on the feature/sounds branch, continuing @Mr-Jay-Gatsby's work.  For discussion up to this point, see https://github.com/freezy/VisualPinball.Engine/pull/453

### DIFF
--- a/VisualPinball.Engine/Common/Constants.cs
+++ b/VisualPinball.Engine/Common/Constants.cs
@@ -160,5 +160,6 @@ namespace VisualPinball.Engine.Common
 		public const string ActionSelfTest = "Self Test";
 		public const string ActionLeftAdvance = "Left Advance";
 		public const string ActionRightAdvance = "Right Advance";
+		public const string ActionExitTable = "Exit Table";
 	}
 }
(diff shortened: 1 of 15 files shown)


linojon:
sorry 'bout all the commits, i like to do things in small increments when possible, and didnt squash the commits before creating the PR. Also the "Q to exit" isnt part of this feature, is a minor change for my convenience that i added to vpe-volley https://github.com/jsm174/vpe-volley/blob/sounds/Assets/Volley/Quitter.cs

freezy:
I've had a quick look at the previous code. About objects that aren't rendered but emit sounds: How about creating a new `SoundEmitterComponent` that implements `ISoundEmitter` and where you're basically able to define the sound triggers by yourself. Each of those triggers would then result in a coil slot that is linkable through the visual scripting (or anything that can connect to a coil, really). 

This way, you'd create your chimes GameObject, add a trigger for each chime, and your `MechSoundsComponent` will show these for you to assign. Then, in visual scripting, hook the score reels to the coils that are exposed.

linojon:
ok, took a couple tries but i've rebased out the extraneous files.

freezy:
Closed by #506.

PR #473, opened 2024-09-06
Fix compiler warnings

arthurkehrwald:
Fixes three compiler warnings.
- 2x CS8618 in LampPLayer.cs
- CS0252 in RampComponent.cs


freezy (approved):
Thanks! ❤️

PR #475, opened 2024-09-06
Log info instead of warning when adding wire

arthurkehrwald:
Adding a dynamic wire should not cause a warning, or at least I don't see why.


freezy (approved):
Yup makes sense.

PR #481, opened 2024-09-20
Fix null error with code-created input actions

arthurkehrwald:
If other code in the Unity project creates input actions that are not part of an action map, a null error is raised every time the associated key binds are pressed by the user.


freezy (approved):
Thanks!

PR #485, opened 2024-10-31
Fix NullReferenceException on subsequent runs

### LINKED ISSUE #472: NullReferenceException on subsequent runs
Any time the Unity editor enters play mode after the initial run, an error message appears.
```
NullReferenceException: Object reference not set to an instance of an object
Unity.Collections.UnsafeQueueBlockPool.OnDomainUnload (System.Object sender, System.EventArgs e) (at ./Library/PackageCache/com.unity.collections@2.1.4/Unity.Collections/UnsafeQueue.cs:139)
System.AppDomain.DoDomainUnload () (at <8ce0bd04a7a04b4b9395538239d3fdd8>:0)
```

arthurkehrwald:
Fixes #472 by initializing the physics engine's event queue in the `Awake` function instead of using a field initializer.


arthurkehrwald:
This fix is included in #486

Issue #488, opened 2024-11-11
Physics engine uses different units for time


arthurkehrwald:
Seconds, milliseconds, and microseconds are all used in the physics code. Often without the variable names indicating what unit is used. This is tedious to work with because every time you work with a time variable you have to figure out what unit it is in. Is there any reason not to just use seconds for everything?

freezy:
Most of the physics code was ported 1:1 from VPX, so much of the nomenclature comes from there. 

So no, there is no particular reason, certainly not deliberate, and yes, a PR to improve this is welcome. :)

PR #489, opened 2024-11-11
Fix bumper animations

arthurkehrwald:
Fixes the bumper animations broken in #483 . To be honest, I don't understand how this was working before. The delta time passed into the animation `Update` functions was the Unity frame time in milliseconds. But this was done in a loop that runs once for every millisecond of frame time. So the animations were just over before they could ever appear on screen. I also found it confusing how the physics engine uses seconds, milliseconds, and microseconds, often without the variable name indicating the unit used. (See #488)


freezy:
Thanks, I'll test this later tonight!

freezy:
This code has already gone upstream, right?

arthurkehrwald:
No idea, I'd also need to check

freezy:
@arthurkehrwald Thanks, I did and it seems to be the case. I'll let you close this one if you confirm as well.

PR #491, opened 2024-11-11
Move bumper impulse into physics loop

arthurkehrwald:
#483 added the impact of the bumper on the ball directly inside the `BumperApi.OnCoil` method. This PR moves the execution of that code into the physics loop using the `PhysicsEngine.ScheduleAction` function.

### DIFF
--- a/VisualPinball.Unity/VisualPinball.Unity/Game/PhysicsEngine.cs
+++ b/VisualPinball.Unity/VisualPinball.Unity/Game/PhysicsEngine.cs
@@ -303,6 +303,19 @@ namespace VisualPinball.Unity
 			}
 		}
 
+		internal PhysicsState CreateState()
+		{
+			var events = _eventQueue.Ref.AsParallelWriter();
+			var env = _physicsEnv.Ref[0];
+			return new PhysicsState(ref env, ref _octree, ref _colliders, ref _kinematicColliders,
+				ref _kinematicCollidersAtIdentity, ref _kinematicTransforms.Ref, ref _updatedKinematicTransforms.Ref,
+				ref _nonTransformableColliderTransforms.Ref, ref _kinematicColliderLookups, ref events,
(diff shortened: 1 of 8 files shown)


arthurkehrwald:
I've added an ugly workaround to fix another issue with #483 I [described in its thread](https://github.com/freezy/VisualPinball.Engine/pull/483#issuecomment-2478681132). I added a flag called `IsSwitchWiredToCoil` to the `BumperState` struct that allows the physics engine to immediately push away balls that enter the bumper collider just like before, but only if the bumper switch is actually wired to the coil. I'm still not happy with this, because it's way more complicated than it should be and it also means that the impulse is almost applied twice: Once when the ball first enters the collider and then again when the coil is triggered by the bumper switch. The only reason this doesn't cause issues is because the ball is already pushed out of the collider by the time the second impulse is applied to all balls inside the collider. We do still need the second impulse though, because it is no longer guaranteed that the coil activation was caused by a ball hitting the bumper, since the game logic engine can activate the bumper coil whenever it wants.

freezy:
Okay, I'll review this the next days.

freezy:
Rebased and adapted this. @arthurkehrwald can you have a quick look, then I'll merge and address #494.

freezy:
Actually, it looks like #494 also went upstream already 🤔

arthurkehrwald:
Your commit just fixes up this PR for a change you made to the signature of Collide3DWall right? I don't have much to say about this. Seems good. In general, I still think this solution isn't great, because it solves a general problem in a way specific to the bumpers. I thought about this some more and I think what we really need is something similar to a hardware rule, but for the physics engine. The game logic engine could translate its hardware rules into something the physics engine can understand so we don't need ugly workarounds like this for game items that need to react immediately in the physics engine without a detour through game logic. We are going to have the same problem with the slingshots again if we also want to make those controllable by the game logic engine.

freezy:
Yes, hardware rule for physics engine makes complete sense for me. I'm merging this for now.

PR #494, opened 2024-11-12
Hardwire bumpers by default

arthurkehrwald:
Makes bumpers 'just work' by default so users who don't want any special game logic for their bumpers don't have to wire them manually. Adds the option to disable the wire to the `BumperComponent` inspector.


freezy:
Can I merge those bumper PRs into my `transformable-colliders` branch? It would avoid having to rebase it again, and I've also changed the bumper code a bit to make it more freely transformable.

arthurkehrwald:
Sure go ahead

freezy:
This already went upstream too, right?

arthurkehrwald:
Yes it did

freezy:
Alright, closing then.

PR #501, opened 2025-01-06
Fix hit events

### LINKED ISSUE #499: Gate does not fire hit events
`GateApi` inherits `CollideableApi`, which determines whether hit events will be fired using the virtual `FireHitEvents` flag. This flag is false by default, but most game items that inherit it override it with true or a value exposed to the user in the inspector. Why does the gate not do that? Seems like an oversight.

arthurkehrwald:
Fixes #499 and fixes the ColliderComponent.HitEvent setting. It was previously ignored.


freezy (approved):
Okay I'll get this in before my other PR.

freezy:
Thanks!

PR #502, opened 2025-01-09
Correctly set isHardwareRule in AddHardwareRule

arthurkehrwald:
When adding hardware rules via `Player.AddHardwareRules` the argument `isHardwareRule` must explicitly be set to true, because it is false by default. This argument is stored in `WireDestConfig.IsHardwareRule` and checked when the hardware rule is removed. If it is not true, the rule is not removed using `Player.RemoveHardwareRule` and when using something like MPF that removes and adds the same rules whenever the ball drains, hardware rules will start piling up. With this PR, hardware rules can be removed.


freezy:
Thank! And sorry for the delay!

PR #504, opened 2025-01-16
Make IGameLogicEngine.Init asynchronous

arthurkehrwald:
I want this so I don't have to block the main thread while connecting the `MpfGameLogicEngine` to MPF


freezy:
This needs to be updated for the PinMAME and Visual Scripting game engines as well, right?

arthurkehrwald:
Yeah I made PRs for that too

PR #506, opened 2025-01-30
Mechanical sound support part 4

### LINKED ISSUE #449: Handle Mechanical Sounds

freezy:
This is @arthurkehrwald's branch rebased, including a few more fixes for #500.

Many thanks to @Mr-Jay-Gatsby, @linojon and @arthurkehrwald for working on this. Please let me all know how you want to be credited in `package.json` and elsewhere. 

This PR closes #449, #453, #466 and #495.


arthurkehrwald:
Glad to see this make its way into the main branch. To credit me, just use my name: Arthur Kehrwald.

PR #515, opened 2025-02-17
Packaging foundation 

freezy:
This PR lays the foundation for a new `.vpe` file format. It serializes a table into a single file and supports re-importing into the editor. The format is also runtime compatible (although we don’t yet have a dedicated player repository, tests have been conducted).

For a more detailed overview, see the [README](https://github.com/freezy/VisualPinball.Engine/blob/feature/packaging/VisualPinball.Unity/VisualPinball.Unity/Packaging/README.md) at the root of the `Packaging` folder.

The two main classes, `PackageWriter` and `PackageReader`, reside in the editor namespace. Another reader for runtime usage is planned.

Still missing is serialization of native components. This is necessary for Visual Scripting, and surely for other packages.

Note that the OpenMcdf implementation is still available, allowing the container format to be switched from ZIP to OLE Storage (as used by VPX). However, as stated in the linked README, there is little benefit to using OLE Storage. ZIP is far more common and provides built-in compression (and encryption, if needed). Also, the storage API is quite awkward, since it was modeled after OpenMcdf's API, so there are optimizations to be done.

Also note that this isn't optimized yet at all. File size is pretty much that raw texture data of your current scene, and lots can be done with image compression. I've done some testing with KTX2 and quantization, but [glTFast](https://docs.unity3d.com/Packages/com.unity.cloud.gltfast@6.0/manual/index.html) doesn't yet support this extension for writing.


freezy:
@arthurkehrwald comments welcome. I'll probably merge this in a day or two.

VisualPinball.Unity/VisualPinball.Unity/Sound/SoundAsset.cs
```diff
@@ -30,43 +31,46 @@ namespace VisualPinball.Unity
...
-		private enum SelectionMethod
+		public enum SelectionMethod
 		{
 			RoundRobin,
 			Random
 		}
 
-		[SerializeField]
-		private string _description;
+		[FormerlySerializedAs("_description")]
```
arthurkehrwald, line 44: I know I'm a bit late, but I noticed this when rebasing. Is this to preserve backwards compatibility? I don't mind breaking these references in my test project. And I don't think anyone has built anything with the audio in the few weeks it's been in here.

freezy: Yes if it's okay to break them, let's remove those.

VisualPinball.Unity/VisualPinball.Unity/Sound/SoundAsset.cs
```diff
@@ -30,43 +31,46 @@ namespace VisualPinball.Unity
...
+		public float FadeOutTime;
 
+		[FormerlySerializedAs("_isSpatial")]
 		[Tooltip("Should the sound appear to come from the position of the emitter?")]
-		[SerializeField]
-		private bool _isSpatial = true;
+		public bool IsSpatial = true;
 
 		[SerializeField]
 		private AudioMixerGroup _audioMixerGroup;
```
arthurkehrwald, line 76: Are you just not packing this reference into the table file at all?

arthurkehrwald: Wait this is a scriptable object. It's not packed manually at all. Are we just using Unity's serialization for those? And why did you make all the fields public? They are not accessed anywhere.

freezy: Yes the entire file is serialized as JSON.

TBH I'm not sure anymore why I made the fields public, possibly because they wouldn't be picked up by the JSON serializer (althought I think there are attributes to explicitly serialize private members).

I'm open to make them private again if it doesn't break the serialization.

PR #521, opened 2025-02-27
Sound upgrades

arthurkehrwald:
There are now three types of sound assets: Sound effects, and the newly added Music and Callout types. Unlike sound effects, there is some coordination required for playing music and callouts, since there should never be multiple songs or multiple callouts playing at once. This is taken care of by the new `MusicCoordinator` and `CalloutCoordinator` components that are automatically added to the table root when importing or creating a new table. Instead of each `SoundComponent` playing its sounds itself, they submit requests to these coordinators. When a request is made, the coordinator returns an id that can be used to cancel it later.

The music requests are arranged in a stack. The most recent, highest priority request is played. When a request is interrupted, it fades out and the new top request fades in if there is any. The fading is programmed in the `Update` loop rather than asynchronously to be able to respond to unpredictable changes. Each song will always fade towards its desired volume. When switching between two requests that have the same music asset, there is no transition at all.

The callout requests are arranged in a queue. New callout requests 'cut in line' in front of all lower priority requests. There is a customizable minimum pause between subsequent callouts.

There is now a default audio mixer that I think all tables should use so we can have global volume settings across all tables. Its audio mixer groups are applied by default to newly created sound assets. It includes a 'Duck' effect that lowers the volume of music and sound effects when a callout is played.

For callouts and sound effects, I have added more detailed customization for what should happen when the sound is triggered again before the previous sound has finished. You can choose to stop the previous sound, fade out the previous sound, don't start a new sound, or play both in parallel.

The `IsSpatial` option was replaced with the `Type` option. Users will now choose whether a sound effect is 'synthetic' or 'mechanical'. I think this is easier to understand. It is only for sound effects, since I don't think music or callouts should ever be spatial, except in VR, but that's a whole new topic.

Spam protection is now done at the sound asset level and not at the component level. This is to catch scenarios like a ball hitting two drop targets at the same time. It just sounds terrible when the same sound is played twice.

I tried to add packing support to the new classes and removed the field attributes for backwards compatibility. @freezy Can you check if it works?

Note: Generally don't test this with any version of Unity 6 before .32. Those versions lie about whether or not the editor window has focus which causes audio playback issues.


arthurkehrwald:
This breaks importing VPX tables. Not ready for merge.

freezy:
Awesome, sorry for reacting so late on this, was working on some more physics bugs.

Will you rebase this, or should I do it? Giving it a spin later.

arthurkehrwald:
I don't think this needs to be rebased. I already merged your packaging changea into this branch.

VisualPinball.Unity/VisualPinball.Unity.Editor/Sound/BinaryEventSoundComponentInspector.cs
```diff
@@ -0,0 +1,37 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace VisualPinball.Unity.Editor
+{
+	[CanEditMultipleObjects]
+	public class BinaryEventSoundComponentInspector : SoundComponentInspector
```
freezy, line 24: I'll rename this to `BinaryEventSoundInspector`, okay?

VisualPinball.Unity/VisualPinball.Unity.Editor/Sound/CalloutAssetInspector.cs
```diff
@@ -0,0 +1,39 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+using UnityEditor;
+using UnityEngine;
+using UnityEngine.UIElements;
+
+namespace VisualPinball.Unity.Editor
+{
+	[CustomEditor(typeof(CalloutAsset)), CanEditMultipleObjects]
+	public class VoiceAssetInspector : SoundAssetInspector
```
freezy, line 24: Okay to rename to `CalloutAssetInspector`? It's less confusing when named the same as the asset.

VisualPinball.Unity/VisualPinball.Unity.Editor/Sound/BinaryEventSoundComponentInspector.cs
arthurkehrwald: Sure

VisualPinball.Unity/VisualPinball.Unity.Editor/Sound/CalloutAssetInspector.cs
arthurkehrwald: Yes I forgot to rename that

VisualPinball.Unity/VisualPinball.Unity/Sound/CalloutRequester.cs
```diff
@@ -0,0 +1,51 @@
...
+// along with this program. If not, see <https://www.gnu.org/licenses/>.
+
+using UnityEngine;
+
+namespace VisualPinball.Unity
+{
+	/// <summary>
+	/// Requests a callout from the callout coordinator when enabled. Intended for testing.
+	/// </summary>
+	public class CalloutRequester : MonoBehaviour
```
freezy, line 24: This one should be packed as well, right?

freezy: Or is it instantiated during runtime?

freezy: Ah, intended for testing.

arthurkehrwald: It is bit really meant to be used in actual tables, but probably best to pack it

freezy: Okay I'll add it.

VisualPinball.Unity/VisualPinball.Unity/Sound/MusicPlayer.cs
```diff
@@ -0,0 +1,129 @@
...
+#endif
+
+namespace VisualPinball.Unity
+{
+	/// <summary>
+	/// Manages the playback of a single music asset according to the <c>ShouldPlay</c> property
+	/// controlled by <c>MusicCoordinator</c>. Always fades towards the desired state, unless
+	/// <c>StartAtFullVolume</c> is true when starting.
+	/// </summary>
+	public class MusicPlayer : MonoBehaviour
```
freezy, line 29: This is instantiated during runtime and doesn't need packaging, right?

freezy:
Thanks, this is awesome. I'v pushed some changes, mostly related to packaging. You need your MonoBehavior to implement `IPackable` in order to get it packed. I've included the two Requeters to the package as well.

I'll change the way referenced files from ScriptableObjects are packed because right now you need to do it manually and I'd like to get this done automatically (currently only sound effect clips are packaged). 

Lastly, it would be great if in the future, you could rebase your branches, it's just a minor thing but it allows for linear history and I don't have to change the repo settings when merging it back to master. ;)

freezy:
Okay, pushed my last changes. If you could validate if everything still works that'd be great, then I'll merge this.

PR #527, opened 2025-03-16
Fix compile error

arthurkehrwald:
Idk if this is right, I just had to get it to compile to work on something else.


freezy:
Yes, sorry about that. I've put `Vertex3D.Zero`, but I'll have to retest this. It's about getting the UVs of surfaces correct.

As far as I've seen, this has been included in #528, it's already merged now?

PR #531, opened 2025-03-25
Document sound features for table authors

arthurkehrwald:


VisualPinball.Unity/Documentation~/creators-guide/manual/sound.md
```diff
@@ -0,0 +1,144 @@
...
+  two drop targets at once.
+
+### Callouts
+
+Callouts are short voice lines that are typically triggered when the player does
+something special, like triggering multiball. The reason for the distinction
+between callouts and sound effects is that callouts tend to be much more
+significant to the player than other sounds. As such, their playback is managed
+by a table-wide _Callout Coordinator_ component at the root of the table
+hierarchy that ensures callouts are played one at a time. When triggering a
```
freezy, line 89: New paragraph?

```suggestion
hierarchy that ensures callouts are played one at a time. 

When triggering a
```

freezy (approved):
Awesome doc, thanks! I've replaced the editor shots with a higher resolution and aligned them to the right side like in the rest of the doc.

Other comments below.

One thing you could also add is doc links in the code, so the ❓-icon of the component opens up the documentation. Example:

```cs
[HelpURL("https://docs.visualpinball.org/creators-guide/manual/sound.html")]
```

PR #532, opened 2025-03-30
Fix: Dispose disposable sound event sources

arthurkehrwald:
`EventSoundComponent` disposes the event source object in `OnDisable` if it implements `IDisposable`


freezy (approved):
Awesome, thanks!

PR #551, opened 2025-11-23
Update installing-vpe.md

OakAsteroid:
Corrected "Add package by name..." to "Install package by name..."


OakAsteroid:
In Unity 6000.2.13f1 on Mac, the + menu in Package Manager says "Install" instead of "Add". 
Package Manager screenshot

